PR #1, opened 2022-05-09
Add initial code

arcusfelis:
Check [readme](https://github.com/esl/cets/blob/rebased/README.md) for description


README.md
```diff
@@ -0,0 +1,55 @@
...
+It exports functions:
+
+- `start(Tab, Opts)` - starts a new table manager.
+   There is one gen_server for each node for each table.
+- `insert(Server, Rec)` - inserts a new object into a table.
+- `insert_many(Server, Records)` - inserts several objects into a table.
+- `delete(Server, Key)` - deletes an object from the table.
+- `delete_many(Server, Keys)` - deletes several objects from the table.
+
+cets_join module contains the merging logic.
```
chrzaszcz:
```suggestion
`cets_join` module contains the merging logic.
```

README.md
```diff
@@ -0,0 +1,55 @@
...
+
+cets_discovery module handles search of new nodes.
```
chrzaszcz:
```suggestion
`cets_discovery` module handles search of new nodes.
```
...and it would be good to format the remainder of this page as well.

src/cets_long.erl
```diff
@@ -0,0 +1,73 @@
...
+    Pid = spawn_mon(Info, Parent, Start),
+    try
+            case Catch of
+                true -> just_run_safely(Info#{what => long_task_failed}, Fun);
+                false -> Fun()
+            end
+        after
+            Diff = diff(Start),
+            ?LOG_INFO(Info#{what => long_task_finished, time_ms => Diff}),
+            Pid ! stop
```
pawlooss1: This indent is unnecessary big. I see a similar case in some other places as well. What do you think about plugging in some code formatter, such as https://github.com/WhatsApp/erlfmt? I think it could help us keep the formatting consistent.

src/cets_long.erl
```diff
@@ -0,0 +1,73 @@
...
+    run(Info, Fun, false).
+
+run(Info, Fun, Catch) ->
+    Parent = self(),
+    Start = erlang:system_time(millisecond),
+    ?LOG_INFO(Info#{what => long_task_started}),
+    Pid = spawn_mon(Info, Parent, Start),
+    try
+            case Catch of
+                true -> just_run_safely(Info#{what => long_task_failed}, Fun);
```
pawlooss1: Putting `what => long_task_failed` in this line suggests that our task has already failed. I think it would be more straightforward if we just put it directly in the logging in line 71.

src/cets_call.erl
```diff
@@ -0,0 +1,109 @@
...
+-export([async_operation/2]).
+-export([sync_operation/2]).
+-export([wait_response/2]).
+
+-type request_id() :: cets:request_id().
+-type op() :: cets:op().
+-type server_ref() :: cets:server_ref().
+-type long_msg() :: cets:long_msg().
+
+%% Do gen_server:call with better error reporting.
```
pawlooss1: It should be `Does` instead of `Do`.

src/cets_call.erl
```diff
@@ -0,0 +1,109 @@
...
+%% It would print a warning if the call takes too long.
```
pawlooss1: I'd change "It would print" to "It will log".

src/cets_call.erl
```diff
@@ -0,0 +1,109 @@
...
+    after Timeout ->
+            erlang:demonitor(Mon, [flush]),
+            error(timeout)
+    end.
+
+%% Wait for response from the remote nodes that the operation is completed.
+%% remote_down is sent by the local server, if the remote server is down.
+wait_for_updated(Mon, {Servers, MonTab}) ->
+    try
+        wait_for_updated2(Mon, Servers)
```
pawlooss1: In MongooseIM I saw a naming convention for such case `do_FunctionName` and I like it. Maybe we could rename this one to follow it? I think that giving number suffixes to function names can be confusing for the reader.

rebar.config
```diff
@@ -0,0 +1,9 @@
+{dialyzer, [
+   {warnings,
+    [no_return, no_unused, no_improper_lists, no_fun_app, no_match,
+     no_opaque, no_fail_call, no_contracts, no_behaviours,
+     no_undefined_callbacks, unmatched_returns, error_handling,
+     underspecs
+     % overspecs, specdiffs
```
pawlooss1: Please either remove or uncomment this line.

src/cets.app.src
```diff
@@ -0,0 +1,9 @@
+{application, cets,
+ [{description, "Clustered Erlang Term Storage"},
+  {vsn, "0.1"},
+  {modules, []},
+  {registered, []},
+  {applications, [kernel, stdlib]},
+  {env, []}
+% {mod, {cets_app, []}
```
pawlooss1: Please either remove or uncomment this line.

src/cets_discovery_file.erl
```diff
@@ -0,0 +1,28 @@
...
+-module(cets_discovery_file).
+-behaviour(cets_discovery).
+-export([init/1, get_nodes/1]).
+
+-include_lib("kernel/include/logger.hrl").
+
+init(Opts) ->
+    Opts.
+
+get_nodes(State = #{disco_file := Filename}) ->
```
pawlooss1, line 23: I think a `spec` here would be useful for further development. Especially that it is a callback for the `cets_discovery` behaviour.

src/cets_join.erl
```diff
@@ -0,0 +1,78 @@
...
+-include_lib("kernel/include/logger.hrl").
+
+%% Adds a node to a cluster.
+%% Writes from other nodes would wait for join completion.
+%% LockKey should be the same on all nodes.
+join(LockKey, Info, LocalPid, RemotePid) when is_pid(LocalPid), is_pid(RemotePid) ->
+    Info2 = Info#{local_pid => LocalPid,
+                  remote_pid => RemotePid, remote_node => node(RemotePid)},
+    F = fun() -> join1(LockKey, Info2, LocalPid, RemotePid) end,
+    cets_long:run_safely(Info2#{what => join_failed}, F).
```
pawlooss1: If I get this correctly, `what` key of this map is going to be overwritten in cets_long.erl in every logging line. Could you check it?

src/cets.erl
```diff
@@ -0,0 +1,420 @@
...
+%%   remove/2).
+-spec start(table_name(), start_opts()) -> {ok, pid()}.
+start(Tab, Opts) when is_atom(Tab) ->
+    gen_server:start({local, Tab}, ?MODULE, {Tab, Opts}, []).
+
+-spec stop(server_ref()) -> ok.
+stop(Server) ->
+    gen_server:stop(Server).
+
+-spec dump(server_ref()) -> Records :: [tuple()].
```
```suggestion
-spec dump(table_name()) -> Records :: [tuple()].
```

test/cets_SUITE.erl
```diff
@@ -0,0 +1,259 @@
...
+init_per_testcase(test_multinode_auto_discovery, Config) ->
+    ct:make_priv_dir(),
+    Config;
+init_per_testcase(_, Config) ->
+    Config.
+ 
+end_per_testcase(_, _Config) ->
+    ok.
+ 
+test_multinode(Config) ->
```
pawlooss1, line 122: This test checks a few thing, e.g. joining nodes (both before and after adding data), inserting values, deleting values, bulk operations. Could you split it to minimal test cases where each one tests exactly one thing?

pawlooss1 (commented):
Nice work 👍 I left some comments that I think could improve readability and maintainability.

run_test.sh
```diff
@@ -0,0 +1,2 @@
+#!/usr/bin/env bash
+./rebar3 ct --sname=ct1
```
chrzaszcz: There is no `rebar3` in this repo. I would just do:

```suggestion
rebar3 ct --sname=ct1
```

src/cets.erl
```diff
@@ -0,0 +1,420 @@
...
+-type handle_down_fun() :: fun((#{remote_pid := pid(), table := table_name()}) -> ok).
+-type start_opts() :: #{handle_down := handle_down_fun()}.
+
+-export_type([request_id/0, op/0, server_ref/0, long_msg/0]).
+
+%% API functions
+
+%% Table and server has the same name
+%% Opts:
+%% - handle_down = fun(#{remote_pid => Pid, table => Tab})
```
```suggestion
%% - handle_down = fun(#{remote_pid := Pid, table := Tab})
```

src/cets_join.erl
chrzaszcz: > If I get this correctly, `what` key of this map is going to be overwritten in cets_long.erl in every logging line. Could you check it?

What's more, this can be misleading especially when debugging/tracing. Finding `what => join_failed` in the arguments would suggest that something has already failed.

src/cets_join.erl
```diff
@@ -0,0 +1,78 @@
...
+    Paused = [{Pid, cets:pause(Pid)} || Pid <- AllPids],
+    %% Merges data from two partitions together.
+    %% Each entry in the table is allowed to be updated by the node that owns
+    %% the key only, so merging is easy.
+    try
+        cets:sync(LocalPid),
+        cets:sync(RemotePid),
+        {ok, LocalDump} = remote_or_local_dump(LocalPid),
+        {ok, RemoteDump} = remote_or_local_dump(RemotePid),
+        RemF = fun(Pid) -> cets:send_dump(Pid, LocPids, LocalDump) end,
```
chrzaszcz, line 73: Whole table sent in one go? It might be ok for now, but I am thinking about the case where we need to send the entire session table from one node to another when it joins the cluster. We would need to allocate a lot of memory on the process heap for this operation. Could you check what happens for a large table?

src/cets_join.erl
```diff
@@ -0,0 +1,78 @@
+%% @doc Cluster join logic.
+-module(cets_join).
+-export([join/4]).
+-include_lib("kernel/include/logger.hrl").
+
+%% Adds a node to a cluster.
+%% Writes from other nodes would wait for join completion.
+%% LockKey should be the same on all nodes.
+join(LockKey, Info, LocalPid, RemotePid) when is_pid(LocalPid), is_pid(RemotePid) ->
```
chrzaszcz, line 12: Please add type specs

Also prevent cets from killing the processes (and losing ETS data) in the event of accidentally calling it with LocalPid same as RemotePid. I checked it and it crashes.

src/cets.erl
```diff
@@ -0,0 +1,420 @@
...
+        false ->
+            erlang:monitor(process, RemotePid),
+            [RemotePid | add_servers2(OtherPids, Servers)];
+        true ->
+            ?LOG_INFO(#{what => already_added,
+                        remote_pid => RemotePid, remote_node => node(RemotePid)}),
+            add_servers2(OtherPids, Servers)
+    end;
+add_servers2([], _Servers) ->
+    [].
```
chrzaszcz: Minor, style: I think it would be a good idea to have a separate function for adding a single server.
```suggestion
add_servers(Pids, Servers) ->
    lists:sort(lists:foldl(fun(Pid) -> add_server(Pid, Servers) end, Servers, Pids)).

add_server(RemotePid, Servers)
  when is_pid(RemotePid), RemotePid =/= self() ->
    case has_remote_pid(RemotePid, Servers) of
        false ->
            erlang:monitor(process, RemotePid),
            [RemotePid | Servers];
        true ->
            ?LOG_INFO(#{what => already_added,
                        remote_pid => RemotePid, remote_node => node(RemotePid)}),
            Servers
    end.
```

src/cets.erl
```diff
@@ -0,0 +1,420 @@
...
+    cets_call:long_call(Server, get_info).
+
+%% gen_server callbacks
+
+-spec init({table_name(), start_opts()}) -> {ok, state()}.
+init({Tab, Opts}) ->
+    process_flag(message_queue_data, off_heap),
+    MonTab = list_to_atom(atom_to_list(Tab) ++ "_mon"),
+    _ = ets:new(Tab, [ordered_set, named_table, public]),
+    _ = ets:new(MonTab, [public, named_table]),
```
chrzaszcz, line 245: This pattern matching does nothing.

src/cets_mon_cleaner.erl
```diff
@@ -0,0 +1,78 @@
...
+
+flush_all_checks() ->
+    receive check -> flush_all_checks() after 0 -> ok end.
+
+start_timer(Interval) ->
+    erlang:send_after(Interval, self(), check).
+
+-spec handle_check(state()) -> state().
+handle_check(State = #{mon_tab := MonTab}) ->
+    check_loop(ets:tab2list(MonTab), MonTab),
```
chrzaszcz, line 79: After you call `cets:stop/1`, the cleaner will crash here with `badarg`.

src/cets_join.erl
```diff
@@ -0,0 +1,78 @@
...
+    Info2 = Info#{local_pid => LocalPid,
+                  remote_pid => RemotePid, remote_node => node(RemotePid)},
```
chrzaszcz: Is there a point in calling them `local` and `remote` if even in tests the first Pid is the remote one? Either ensure the first one is local, or rename them.

src/cets_long.erl
```diff
@@ -0,0 +1,73 @@
...
+-module(cets_long).
+-export([run_spawn/2, run/2, run_safely/2]).
+
+-include_lib("kernel/include/logger.hrl").
+
+%% Spawn a new process to do some memory-intensive task
+%% This allows to reduce GC on the parent process
+%% Wait for function to finish
+%% Handles errors
+run_spawn(Info, F) ->
```
chrzaszcz, line 20: Please add specs for exported functions.

src/cets_join.erl
chrzaszcz: Additionally, there will be more `cets_long` calls inside this `cets_long` call. I don't think we need doubled monitors. Maybe skip the outer `cets_long` entirely and just catch errors?

src/cets_join.erl
```diff
@@ -0,0 +1,78 @@
...
+    AllPids = LocPids ++ RemPids,
+    Paused = [{Pid, cets:pause(Pid)} || Pid <- AllPids],
+    %% Merges data from two partitions together.
+    %% Each entry in the table is allowed to be updated by the node that owns
+    %% the key only, so merging is easy.
+    try
+        cets:sync(LocalPid),
+        cets:sync(RemotePid),
+        {ok, LocalDump} = remote_or_local_dump(LocalPid),
+        {ok, RemoteDump} = remote_or_local_dump(RemotePid),
```
chrzaszcz, line 72: Here the local vs remote semantics gets particularly blurry. I mean: the Local one can be a remote one instead, it is so in the tests.

src/cets_mon_cleaner.erl
```diff
@@ -0,0 +1,78 @@
...
+
+terminate(_Reason, _State) ->
+    ok.
+
+code_change(_OldVsn, State, _Extra) ->
+    {ok, State}.
+
+-spec schedule_check(state()) -> state().
+schedule_check(State = #{interval := Interval, timer_ref := OldRef}) ->
+    _ = erlang:cancel_timer(OldRef),
```
chrzaszcz, line 64: What is the point of pattern matching to anything?

.github/workflows/ci.yml
```diff
@@ -0,0 +1,50 @@
...
+
+permissions:
+  contents: read
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+
+    container:
+      image: erlang:23
```
chrzaszcz: Update Erlang version.

.gitignore
```diff
@@ -0,0 +1,4 @@
+_build/
+rebar3
```
chrzaszcz: There is no `rebar3` in the repo.

src/cets.erl
```diff
@@ -0,0 +1,420 @@
...
+handle_op(From = {Mon, Pid}, Msg, State) when is_pid(Pid) ->
+    do_op(Msg, State),
+    WaitInfo = replicate(From, Msg, State),
+    Pid ! {cets_reply, Mon, WaitInfo},
+    ok.
+
+replicate(From, Msg, #{mon_tab := MonTab, other_servers := Servers}) ->
+    %% Reply would be routed directly to FromPid
+    Msg2 = {remote_op, From, Msg},
+    replicate2(Servers, Msg2),
```
chrzaszcz: I don't think we gain anything by the increased amount of code and the additional function with a `2` suffix.  `lists:foreach` is also fine for me - easier to read, debug, understand...

```suggestion
    [send_to_remote(RemotePid, Msg2) || RemotePid <- Servers],
```

src/cets.erl
```diff
@@ -0,0 +1,420 @@
...
+    {Servers, MonTab}.
+
+replicate2([RemotePid | Servers], Msg) ->
+    send_to_remote(RemotePid, Msg),
+    replicate2(Servers, Msg);
+replicate2([], _Msg) ->
+    ok.
+
+apply_backlog(State = #{backlog := Backlog}) ->
+    apply_backlog_ops(lists:reverse(Backlog), State),
```
chrzaszcz: Same as above.

src/cets_discovery.erl
```diff
@@ -0,0 +1,146 @@
...
+        tables := [atom()],
+        backend_module := module(),
+        backend_state := state(),
+        timer_ref := reference() | undefined
+       }.
+
+-callback init(map()) -> backend_state().
+-callback get_nodes(backend_state()) -> {get_nodes_result(), backend_state()}.
+
+start(Opts) ->
```
chrzaszcz, line 43: Please add specs to the API functions. `Opts` need a spec.

src/cets_discovery.erl
```diff
@@ -0,0 +1,146 @@
...
+    report_results(Results, State),
+    State#{results := Results}.
+
+schedule_check(State) ->
+    cancel_old_timer(State),
+    TimerRef = erlang:send_after(5000, self(), check),
+    State#{timer_ref := TimerRef}.
+
+cancel_old_timer(#{timer_ref := OldRef}) when is_reference(OldRef) ->
+    _ = erlang:cancel_timer(OldRef),
```
chrzaszcz, line 143: What is the point of this pattern match? It has no effect.

chrzaszcz (commented):
It looks promising 🙂 I added some comments.

src/cets.erl
```diff
@@ -0,0 +1,420 @@
...
+info(Server) ->
+    cets_call:long_call(Server, get_info).
+
+%% gen_server callbacks
+
+-spec init({table_name(), start_opts()}) -> {ok, state()}.
+init({Tab, Opts}) ->
+    process_flag(message_queue_data, off_heap),
+    MonTab = list_to_atom(atom_to_list(Tab) ++ "_mon"),
+    _ = ets:new(Tab, [ordered_set, named_table, public]),
```
chrzaszcz, line 244: Another note: why `ordered_set`? It slows down read time. Also: how about allowing some customisation instead of hardcoding the table type?

arcusfelis: ordered_set is required for ejabberd_sm to work - for reads all sessions of the user, for example.

src/cets.erl
arcusfelis: it prevents from dialyzer warning
```
src/cets.erl
Line 241 Column 5: Expression produces a value of type atom() | ets:tid(), but this value is unmatched
Line 242 Column 5: Expression produces a value of type atom() | ets:tid(), but this value is unmatched
===> Warnings written to _build/default/25.0.1.dialyzer_warnings
```

src/cets_discovery.erl
arcusfelis: prevents from dialyzer warning.

src/cets_join.erl
arcusfelis: Yes, in tests both pids could be local.
But once discovery module calls `join`, Local and Remote pids would be like that.
I could call it Server1 and Server2, but it would be harder to read (probably).

But basically, there is no difference between these two servers, exchange is bi-directional. They both send dumps to each other.

src/cets_join.erl
arcusfelis: There would be a spike of memory usage (probably we need 2 times of the ETS table size).
For the real life ejabberd_sm tables - would be sent pretty fast.

Though, it is the fastest way to send. Also, could cause distributed connection processes to stuck for a second. 

I was thinking if we need pagination to be added - it could be done separately (separate PR), because it would be easier to review the main PR.

src/cets_long.erl
arcusfelis: I have so many questions to erlfmt :) 
At least for nodejs formatters you can configure the style.

That aligning when `when` is used is kinda wtf:

```erlang

resolve(#amp_strategy{deliver = [Value | _]}, #amp_rule{
    condition = deliver, value = Value, action = Action
}) when
    Action == drop orelse Action == error
->
    match;
resolve(#amp_strategy{'match-resource' = Value}, #amp_rule{
    condition = 'match-resource', value = any
}) when
    Value /= undefined
->
    match;
resolve(#amp_strategy{'match-resource' = Value}, #amp_rule{
    condition = 'match-resource', value = Value
}) ->
```

And brackets:

```erlang
    CommandTags = ets:match(
        ejabberd_commands,
        #ejabberd_commands{
            name = '$1',
            tags = '$2',
            _ = '_'
        }
    ),
```

Maybe we should use it in mongooseim ;)

src/cets_join.erl
arcusfelis: it is primarily for _logging_, otherwise we would have something like `node1` and `node2`, `pid1` and `pid2`.

src/cets_mon_cleaner.erl
arcusfelis: for dialyzer, added as a comment

src/cets_mon_cleaner.erl
arcusfelis: added gen_server:stop into terminate of the main server.

test/cets_SUITE.erl
arcusfelis: I added more tests for each operation

src/cets_discovery.erl
chrzaszcz: Which option triggers this warning? I see a lot of them enabled. Sometimes it might be better to disable a check if it forces us to write unnecessary code.

src/cets_join.erl
chrzaszcz: Actually in some tests the local one is the remote one, they are **swapped**, that's what I meant.

test/cets_SUITE.erl
chrzaszcz: Ok, but it just doesn't help with the readability of this particular one.

test/cets_SUITE.erl
```diff
@@ -0,0 +1,359 @@
...
+
+test_multinode(Config) ->
+    Node1 = node(),
+    [Node2, Node3, Node4] = proplists:get_value(nodes, Config),
+    Tab = tab1,
+    {ok, Pid1} = start(Node1, Tab),
+    {ok, Pid2} = start(Node2, Tab),
+    {ok, Pid3} = start(Node3, Tab),
+    {ok, Pid4} = start(Node4, Tab),
+    ok = join(Node1, Tab, Pid3, Pid1),
```
chrzaszcz: This happens a few times: the first pid is remote, and the second is local, while in the `join` function it is the other way around. Maybe don't swap the order to make it easier to understand?

chrzaszcz (approved):
Looks good :+1:

PR #3, opened 2023-06-24
Add bags support and delete_object

arcusfelis:
Changes include:
- Allow to specify the table type.
- Allow to delete an object (or many).

### DIFF
--- a/src/cets.erl
+++ b/src/cets.erl
@@ -126,7 +126,7 @@
 }.
 
 -type handle_down_fun() :: fun((#{remote_pid := pid(), table := table_name()}) -> ok).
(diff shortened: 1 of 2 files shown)


src/cets.erl
```diff
@@ -24,6 +24,8 @@
     insert_many/2,
     delete/2,
     delete_many/2,
+    delete_object/2,
+    delete_object_many/2,
```
chrzaszcz: Please, please name it just `delete_objects` or `delete_many_objects` if you want to keep `many` in the name.

chrzaszcz (commented):
Looks good, just one small request from me.

src/cets.erl
arcusfelis: done

PR #4, opened 2023-07-07
Conflict handler option

arcusfelis:
As a developer I want to use non-random and non-node-specific keys. This means duplicates are possible when two nodes are joined together.

Changes:
- Add an option `handle_conflict` to define a function which could be used to choose between two records if there is a conflict on the cluster join.
- Tests.

### DIFF
--- a/README.md
+++ b/README.md
@@ -30,6 +30,10 @@ You can provide the `handle_conflict` function, which would be called if two rec
 the same key when clustering (just be aware, this function would not protect you against
 overwrites by the same node).
 
+Without `handle_conflict`, conflicting records would overwrite each other when joining.
+When joining nodes we insert all data from one node to another and vice versa without any extra check.
+Check `join_works_with_existing_data_with_conflicts` test case as an example.
+
 # API
 
 The main module is cets.
--- a/README.md
+++ b/README.md
@@ -23,6 +23,13 @@ should be inside an ETS key for this to work (or a pid).
 When some node is down, we remove all records that are owned by this node.
 When a node reappears, the records are added back.
 
+You can also use totally random keys. In this case all keys should be unique.
+We also support tables with type=bag. In such case all records would remain in the table.
+
+You can provide the `handle_conflict` function, which would be called if two records have
+the same key when clustering (just be aware, this function would not protect you against
+overwrites by the same node).
+
 # API
 
 The main module is cets.
--- a/src/cets.erl
+++ b/src/cets.erl
@@ -129,7 +129,9 @@
 -type handle_down_fun() :: fun((#{remote_pid := pid(), table := table_name()}) -> ok).
 -type handle_conflict_fun() :: fun((tuple(), tuple()) -> tuple()).
 -type start_opts() :: #{
-    handle_down => handle_down_fun(), type => ordered_set | bag, keypos => non_neg_integer(),
+    type => ordered_set | bag,
+    keypos => non_neg_integer(),
+    handle_down => handle_down_fun(),
     handle_conflict => handle_conflict_fun()
 }.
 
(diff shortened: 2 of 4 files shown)


src/cets.erl
```diff
@@ -142,9 +147,21 @@
 %%   to make a new async process if you need to update).
 %%   i.e. any functions that replicate changes are not allowed (i.e. insert/2,
 %%   remove/2).
+%% - handle_conflict = fun(Record1, Record2) -> NewRecord
+%%   Called when two records have the same key when clustering.
+%%   NewRecord would be the record CETS would keep in the table under the key.
+%%   Does not work for bags.
+%%   We recomment to define that function if keys could have conflicts.
```
```suggestion
%%   We recommend to define that function if keys could have conflicts.
```

src/cets_join.erl
```diff
@@ -86,3 +88,44 @@ remote_or_local_dump(Pid) when node(Pid) =:= node() ->
...
+    LKey = element(Pos, L),
+    RKey = element(Pos, R),
+    if
+        LKey =:= RKey ->
+            New = F(L, R),
+            apply_resolver_for_sorted(LocalDump, RemoteDump, F, Pos, [New | LocalAcc], [
+                New | RemoteAcc
+            ]);
+        LKey < RKey ->
+            %% Record exist only in the local dump
```
chrzaszcz:
```suggestion
            %% Record exists only in the local dump
```

src/cets_join.erl
```diff
@@ -86,3 +88,44 @@ remote_or_local_dump(Pid) when node(Pid) =:= node() ->
...
+            apply_resolver_for_sorted(LocalDump, RemoteDumpFull, F, Pos, [L | LocalAcc], RemoteAcc);
+        true ->
+            %% Record exist only in the remote dump
```
chrzaszcz:
```suggestion
            %% Record exists only in the remote dump
```

src/cets.erl
```diff
@@ -527,3 +552,13 @@ call_user_handle_down(RemotePid, _State = #{tab := Tab, opts := Opts}) ->
         _ ->
             ok
     end.
+
+-type start_error() :: bag_with_conflict_handler.
+-spec check_opts(start_opts()) -> [start_error()].
+check_opts(Opts) ->
+    check_bag_with_conflict_handler(Opts).
```
chrzaszcz: Minor, style: this extra function is not needed, I would put the logic from lines 561-564 directly here. The clause in line 561 is just as informative as the function name.

README.md
```diff
@@ -23,6 +23,13 @@ should be inside an ETS key for this to work (or a pid).
 When some node is down, we remove all records that are owned by this node.
 When a node reappears, the records are added back.
 
+You can also use totally random keys. In this case all keys should be unique.
+We also support tables with type=bag. In such case all records would remain in the table.
+
+You can provide the `handle_conflict` function, which would be called if two records have
```
chrzaszcz, line 29: Is there a any information what happens when there is no `handle_conflict` function?

chrzaszcz (commented):
Looks good in general, I added minor comments.

src/cets.erl
arcusfelis: could be.
I just was thinking to add more check_* parameters later, if needed so it could be:

```
check_opts(Opts) ->
    check_bag_with_conflict_handler(Opts)
    ++ check_some_other_option(Opts)
    ...
```

README.md
arcusfelis: only in tests, could write though.
Basically, if we have node A and B, data on B would be overwritten with data from A and vice versa.

src/cets.erl
arcusfelis: ok, replaced

chrzaszcz (approved):

PR #6, opened 2023-07-25
Remove mon_tab, use monitors to wait for replies

arcusfelis:
Addresses MIM-1960

We used mon_tab table to notify callers if the remote node is down. So, they don't need to wait for a reply, if there would be no reply.
But it means an extra ETS write and remote for each operation.
So, we just use process_info to get a list of processes, that we could notify about remote node down events.

Sadly, `monitored_by` does not show us references (only pids). It also shows us processes which tried to use `gen_server:call`. We could live with these limitations though.

Tested with https://github.com/arcusfelis/cets_bench/blob/main/samples/run_defaults.exs

Before:

```
Name               ips        average  deviation         median         99th %
mnesia         21.94 K       45.58 μs    ±27.44%       42.26 μs      105.70 μs
cets           20.81 K       48.05 μs    ±25.04%       45.05 μs      100.07 μs
```

After:

```
Name             ips        average  deviation         median         99th %
cets         21.99 K       45.47 μs    ±29.44%       41.44 μs      104.73 μs
mnesia       21.79 K       45.88 μs    ±25.24%       42.08 μs       92.82 μs
```

Changes:
- No mon_tab table.
- Set async_dist=true, the option looks interesting. Removed `nosuspend` (nosuspend does not send a message if distport is busy, so - unreliable message passing with it).
- Use message send instead of cast to send `op` messages. Also, send message to the pid not name.


src/cets.erl
```diff
@@ -397,14 +404,17 @@ handle_down2(_Mon, RemotePid, State = #{other_servers := Servers, mon_tab := Mon
...
-
-notify_remote_down_loop(RemotePid, [{Mon, Pid} | List]) ->
-    Pid ! {remote_down, Mon, RemotePid},
-    notify_remote_down_loop(RemotePid, List);
-notify_remote_down_loop(_RemotePid, []) ->
+notify_remote_down(Mon, RemotePid) ->
+    %% Sadly, we have access to Pids only, without monitor refs.
+    {monitored_by, Pids} = erlang:process_info(self(), monitored_by),
+    %% We don't have monitor reference info
+    Server = self(),
```
NelsonVides: If you're already capturing the variable, just use it 😁 
```suggestion
    Server = self(),
    {monitored_by, Pids} = erlang:process_info(Self, monitored_by),
    %% We don't have monitor reference info
```

src/cets.erl
```diff
@@ -446,11 +456,11 @@ ets_delete_objects(_Tab, []) ->
     ok.
 
 has_remote_pid(RemotePid, Servers) ->
-    lists:member(RemotePid, Servers).
+    lists:keymember(RemotePid, 1, Servers).
 
 reply_updated({Mon, Pid}) ->
-    %% We really don't wanna block this process
-    erlang:send(Pid, {updated, Mon, self()}, [noconnect, nosuspend]).
+    Pid ! {cets_updated, Mon, self()},
```
NelsonVides: Why did we drop the `[noconnect, nosuspend]` options here?

arcusfelis: they both would not sent a message and just return quickly (if detected busy dist port or require connect).
We could use `[noconnect, nosuspend]` but something like:

```erlang
case erlang:send(Pid, M, [noconnect, nosuspend]) of
    ok -> ok;
    Other -> spawn(fun() -> Pid ! M end) % Message is not sent if function returns not ok
end.
```

`noconnect` - pretty useless, we have nodes connected at this point. Especially useless with newer `async_dist=true`. I am not sure about the behaviour when node disappears and we have still messages addressed to that node... could keep `noconnect` probably :D

arcusfelis:
A bit too hacky solution, decided to use aliases to ensure we handle netsplits more reliably (and a disconnected node would not start writing to us again without a proper join)

PR #7, opened 2023-07-27
Use aliases for communication

arcusfelis:
Changes for MIM-1960.

Concurrency fixes:
- Make joining more reliable. We disconnect from the servers which failed to apply the same dump.
- Assign integers to servers when joining. We could do it, because join is done when all nodes could talk to each other.
- Use bitmask when waiting for remote responses - Marginally faster, less garbage and we don't need to return a lot of pids from server.
- Add tests for joining. Use step function to simulate errors.
- **Use process aliases** everywhere.
- Use a separate process instead of mon_tab. Apparently, it allows to write faster (i.e. avoids ETS locks). And we don't really need all ETS table features there, we only wanna store monitors "just in case" one of the remote servers dies. We could store it inside our server, but it makes server to do more work = less throughput. On another hand, sending it to a separate process allows to use another core to handle maps:put/maps:delete operations. It also allows us to call servers remotely (i.e. `cets:insert(Pid, {1})`, where Pid is on the remote node.). Because removing a monitor is a message send operation, not `ets:delete`.
- User would not receive a delayed message after an operation timeout.
- Use standard `gen_server:call` to send requests to the local server. After that `cets_ack` would send us a reply, after all nodes in the cluster ack the operation.

Fix case when:
- join crashes/netsplits and only some servers have applied send_dump. I.e. the servers which applied send_dump would send remote_ops one-way.

Tweaks:
- don't use nosuspend, because it would not send a message if the dist socket is busy. We use async_dist=true instead. Otherwise we could use nosuspend, BUT with `spawn(fun() -> Pid ! Msg end)` if ok is not returned.
- `write_concurrency = true` for mon table - not that much difference though.
- `read_concurrency = true` for the data tables.

We do faster for when testing in parallel (16 benchee workers, 3 nodes):

```
Name             ips        average  deviation         median         99th %
cets          5.99 K      167.00 μs    ±24.20%      162.77 μs      280.81 μs
mnesia        4.80 K      208.26 μs    ±25.27%      198.50 μs      367.14 μs
```

Instead of when having no concurrency (1 benchee worker, 2 nodes):

```
mnesia       21.99 K       45.48 μs    ±28.28%       42.08 μs      108.96 μs
ets          21.89 K       45.68 μs    ±77.62%       43.31 μs       89.36 μs
cets         18.82 K       53.14 μs    ±27.83%       51.80 μs      105.80 μs
```

I.e. no concurrency is consistently 10-20% slower because we don't need to send extra messages to cets_ack and reply is returned directly. Though once we go to the real life scenario with at least 16 workers trying to write into the table, we start seeing lock contention in mnesia. (so, mnesia is a bit slower).

Still prefer having cets_ack to collect replies from the remote nodes, because we don't wanna do any work in the client process (i.e. we just keep gen_server:call api and all the distribution logic is outside of the caller process). If the process has non-empty queue, receiving messages would be probably slower even with selective receives (i.e. monitors) optimisation.

Also, cets_ack process is fast enough, i.e. adding a pool of cets_ack processes does not increase thoughput.

Code in the main branch tested (1 benchee worker, 2 nodes):

```
Name             ips        average  deviation         median         99th %
mnesia       22.37 K       44.70 μs    ±25.34%       41.97 μs      106.20 μs
ets          21.77 K       45.93 μs    ±24.87%       44.29 μs       84.56 μs
cets         19.24 K       51.97 μs    ±21.80%       50.49 μs       99.66 μs
```
Code in the main branch tested (16 benchee workers, 3 nodes):

```
Name             ips        average  deviation         median         99th %
cets          5.55 K      180.19 μs    ±29.33%      173.06 μs      342.94 μs
mnesia        5.14 K      194.54 μs    ±26.85%      187.34 μs      350.60 μs
```

(code used to test https://github.com/arcusfelis/cets_bench - careful, these tests have bad consistency, with 5-10% variation in ips between runs)

**Question: do we really need aliases to communicate between nodes?**
- Actually, just adding DumpRef value (last_applied_dump_ref in the state, assigned at join) to all remote_op-s and comparing it when receiving a remote op would filter out unwanted operations from the nodes, which failed to join. Though, it would not filter out messages from a node which disappeared, was cleaned after, and reappeared and still sends remote_op-s which are not wanted, but would have a valid DumpRef.
- With aliases unwanted remote_op-s would not be even copied into a message box, so it could help not to overload the process during a wild netsplit situation.
- We could check that: dump_ref is matching in the remote_ops and remote_op is from a pid from other_servers list. This means that once we cleanup after a node, we don't receive remote_ops even if node is back (unless it does the proper cets_join again, which would update dump_ref).

**Bit logic is cool idea but probably list operations would work as well**

# This PR should be split into two
- cets_mon_cleaner is replaced with cets_ack. gen_server:call API is used for operations. cets_bits magic is replaced with `lists:delete(RemotePid, OtherPids)`.
- cets_join tests. Aliases are cool, but this means join logic and cets state are more confusing = means more hidden bugs. Probably just setting dump_ref during the join and checking it is fine (and don't allow remote_ops after we handled a netsplit `DOWN` signal from a remote node).

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,6 @@
 _build/
 # vim temporary files
-*.sw*
+.*.sw*
+.*.sv*
+rebar3.crashdump
+codecov.json
--- a/rebar.config
+++ b/rebar.config
@@ -23,7 +23,9 @@
 {profiles, [
     {test, [
         {plugins, [
-            {rebar3_codecov, "0.4.0"}
-        ]}
+            {rebar3_codecov, "0.5.0"}
+        ]},
+        %% Include test suites into the cover report
+        {src_dirs, ["src", "test"]}
     ]}
 ]}.
--- a/src/cets.erl
+++ b/src/cets.erl
@@ -510,7 +510,7 @@ ets_delete_objects(Tab, Objects) ->
 
 reply_updated(Alias, ReplyTo, #{server_mask := Mask}) ->
     %% nosuspend makes message sending unreliable
-    erlang:send(ReplyTo, {cets_updated, Alias, Mask}, [noconnect]).
+    erlang:send(ReplyTo, {ack, Alias, Mask}, [noconnect]).
 
 send_to_remote(RemoteAlias, Msg) ->
     erlang:send(RemoteAlias, Msg, [noconnect]).
--- a/src/cets.erl
+++ b/src/cets.erl
@@ -31,7 +31,7 @@
     table_name/1,
     other_nodes/1,
     other_pids/1,
-    make_alias_for/2,
+    make_aliases_for/2,
     pause/1,
     unpause/2,
     sync/1,
@@ -151,7 +151,7 @@
     | table_name
     | get_info
     | other_pids
-    | {make_alias_for, [pid()]}
+    | {make_aliases_for, [pid()]}
     | {unpause, reference()}
     | send_dump_msg()
     | {apply_dump, DumpRef :: reference()}.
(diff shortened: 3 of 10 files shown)


arcusfelis:
Github is a bit buggy today, so it does not see new commits and does not trigger CI.

arcusfelis:
will be split into two PRs

PR #10, opened 2023-07-31
Add cets:insert_new operation

arcusfelis:
This operation uses one node in the cluster to verify that the inserted record is new.

# Reasoning

It is needed for mod_muc, because it tries to use room-name as a key.
There is a chance the room would be registered twice, so a transaction would be needed.
But insert-new looks like more simple solution (i.e. no need to invent locks for transaction).

# How it works

- we choose one node in the cluster where we would send an initial request.
- this node calls ets:insert_new and replicates the changes. 
- if ets:insert_new returns false - it does not replicate the changes and replies to the caller with `false` instead of `true`.

- We need to call `rpc:async_call(Node, ets, delete, [MonTab, Mon]);` to remove reference from the mon_tab... 

There is not much difference between insert and insert_new (2 nodes, 1 thread to test):

```
Name               ips        average  deviation         median         99th %
mnesia         21.90 K       45.65 μs    ±24.98%       42.80 μs      106.23 μs
ets            20.95 K       47.73 μs    ±76.92%       45.07 μs      100.62 μs
cets           19.65 K       50.89 μs    ±21.38%       50.28 μs       99.67 μs
cets_new       18.96 K       52.74 μs    ±29.13%       51.81 μs      101.72 μs
```

### DIFF
--- a/src/cets_call.erl
+++ b/src/cets_call.erl
@@ -8,6 +8,9 @@
 -export([async_operation/2]).
 -export([sync_operation/2]).
 -export([wait_response/2]).
+-export([send_leader_op/2]).
+
+-include_lib("kernel/include/logger.hrl").
 
 -type request_id() :: cets:request_id().
 -type op() :: cets:op().
@@ -57,20 +60,23 @@ async_operation(Server, Msg) ->
             Mon
     end.
 
--spec sync_operation(server_ref(), op()) -> ok.
+-spec sync_operation(server_ref(), op()) -> ok | Result :: term().
 sync_operation(Server, Msg) ->
     Mon = async_operation(Server, Msg),
     %% We monitor the local server until the response from all servers is collected.
     wait_response(Mon, infinity).
 
 %% This function must be called to receive the result of the multinode operation.
--spec wait_response(request_id(), non_neg_integer() | infinity) -> ok.
+-spec wait_response(request_id(), non_neg_integer() | infinity) -> ok | Result :: term().
 wait_response(Mon, Timeout) ->
     receive
         {'DOWN', Mon, process, _Pid, Reason} ->
             error({cets_down, Reason});
         {cets_reply, Mon, WaitInfo} ->
-            wait_for_updated(Mon, WaitInfo)
+            wait_for_updated(Mon, WaitInfo);
+        {cets_reply, Mon, WaitInfo, Result} ->
+            wait_for_updated(Mon, WaitInfo),
+            Result
     after Timeout ->
         erlang:demonitor(Mon, [flush]),
         error(timeout)
@@ -84,7 +90,10 @@ wait_for_updated(Mon, {Servers, MonTab}) ->
     after
         erlang:demonitor(Mon, [flush]),
         delete_from_mon_tab(MonTab, Mon)
-    end.
+    end;
+wait_for_updated(Mon, false) ->
+    erlang:demonitor(Mon, [flush]), %% not replicated
+    ok.
 
 %% Edgecase: special treatment if Server is on the remote node
 delete_from_mon_tab({remote, Node, MonTab}, Mon) ->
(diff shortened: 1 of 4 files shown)


src/cets_call.erl
```diff
@@ -83,8 +89,18 @@ wait_for_updated(Mon, {Servers, MonTab}) ->
...
-    end.
+        delete_from_mon_tab(MonTab, Mon)
+    end;
+wait_for_updated(Mon, false) ->
+    %% not replicated
+    erlang:demonitor(Mon, [flush]),
+    ok.
+
+%% Edgecase: special treatment if Server is on the remote node
+delete_from_mon_tab({remote, Node, MonTab}, Mon) ->
```
chrzaszcz, line 102: It would be nice to have specs for the `{remote, Node, MonTab}` especially that this tuple is called just MonTab above, which is confusing.

src/cets_call.erl
```diff
@@ -110,3 +126,24 @@ where(Name) when is_atom(Name) -> whereis(Name);
...
+
+%% Sends all requests to a single node in the cluster
+-spec send_leader_op(server_ref(), op()) -> term().
+send_leader_op(Server, Op) ->
+    Leader = cets:get_leader(Server),
+    Res = cets_call:sync_operation(Leader, {leader_op, Op}),
+    case Res of
+        {error, wrong_leader} ->
+            ?LOG_WARNING(#{what => wrong_leader, server => Server, operation => Op}),
+            timer:sleep(10),
```
chrzaszcz: From what I see, this might introduce super-long delays to the code using this function. It doesn't look good.

test/cets_SUITE.erl
```diff
@@ -99,6 +108,141 @@ inserted_records_could_be_read_back_from_replicated_table(_Config) ->
...
+
+insert_new_is_retried_when_leader_is_reelected(_Config) ->
+    Me = self(),
+    F = fun(X) -> Me ! {wrong_leader_detected, X} end,
+    {ok, Pid1} = cets:start(newins1tab_back2, #{}),
+    {ok, Pid2} = cets:start(newins2tab_back2, #{handle_wrong_leader => F}),
+    ok = cets_join:join(join_lock1_insnew_back2, #{}, Pid1, Pid2),
+    Leader = cets:get_leader(Pid1),
+    cets:set_leader(Leader, false),
+    spawn(fun() ->
```
chrzaszcz, line 201: There is no pause, so `insert_new` will be _most likely_ executed before this code... this makes me wonder why we need this function at all.

src/cets_call.erl
arcusfelis: it is 10 milliseconds delay. But I could retry without delay, it will just make our local server less happy during the join operation (and extra load for the remote server too). Alternatively, I can wait for something like "join finished" event.

Will apply retry without delay, add a smarter solution in a separate PR.

test/cets_SUITE.erl
arcusfelis: oh, will add the comment.
Basically, in previous line we use `cets:set_leader(Leader, false)` which makes the call `cets:insert_new` never returning (it retries insert to the leader, but leader says "I am not a leader").

And this repeats until we actually execute code inside spawn function to fix the leader.

src/cets_call.erl
chrzaszcz: I misread as seconds, my bad. 
The only issue is that a bug could cause it to spin forever (especially with zero delay as in the latest commit), at least in theory. A retry counter (or a backoff) would suffice.

src/cets_call.erl
```diff
@@ -110,3 +128,36 @@ where(Name) when is_atom(Name) -> whereis(Name);
...
+    send_leader_op(Server, Op, backoff_intervals()).
+
+send_leader_op(Server, Op, Backoff) ->
+    Leader = cets:get_leader(Server),
+    Res = cets_call:sync_operation(Leader, {leader_op, Op}),
+    case Res of
+        {error, wrong_leader} ->
+            ?LOG_WARNING(#{what => wrong_leader, server => Server, operation => Op}),
+            %% We are free to retry
+            %% While it is infinite retries, the leader election logic is simple.
```
chrzaszcz: It isn't infinite retries anymore

src/cets_call.erl
```diff
@@ -110,3 +128,36 @@ where(Name) when is_atom(Name) -> whereis(Name);
...
+send_leader_op(Server, Op) ->
+    send_leader_op(Server, Op, backoff_intervals()).
+
+send_leader_op(Server, Op, Backoff) ->
+    Leader = cets:get_leader(Server),
+    Res = cets_call:sync_operation(Leader, {leader_op, Op}),
+    case Res of
+        {error, wrong_leader} ->
+            ?LOG_WARNING(#{what => wrong_leader, server => Server, operation => Op}),
+            %% We are free to retry
```
chrzaszcz: For me this comment line is redundant.

chrzaszcz (approved):
Looks good :+1:

PR #11, opened 2023-08-01
Add gradualizer

arcusfelis:
Added gradualizer, tweaked types a bit to pass the checks.
Mainly we have to constrain the result from a function before returning it (if we don't use the same return type)

It also adds more dialyzer checks (which produce similar warnings).

MIM-1981

### DIFF
--- a/src/cets.erl
+++ b/src/cets.erl
@@ -87,10 +87,7 @@
 -type server_pid() :: pid().
 -type server_ref() ::
     server_pid()
-    | atom()
-    | {local, atom()}
-    | {global, term()}
-    | {via, module(), term()}.
+    | atom().
 -type request_id() :: gen_server:request_id().
 -type from() :: gen_server:from().
 -type join_ref() :: cets_join:join_ref().
@@ -667,17 +664,18 @@ handle_get_info(
     #{
         table => Tab,
         nodes => lists:usort(pids_to_nodes([self() | Servers])),
-        size => returns_non_neg_integer(ets:info(Tab, size)),
-        memory => returns_non_neg_integer(ets:info(Tab, memory)),
+        size => assert_type_non_neg_integer(ets:info(Tab, size)),
+        memory => assert_type_non_neg_integer(ets:info(Tab, memory)),
         ack_pid => AckPid,
         join_ref => JoinRef,
         opts => Opts
     }.
 
-%% We are sure that the ETS table exists when running handle_get_info
-%% But gradualizer does not know that, so let us tell it with the spec
--spec returns_non_neg_integer(any()) -> non_neg_integer().
-returns_non_neg_integer(X) -> X.
+%% Assert for gradualizer
+%% ets:info would not return undefined in handle_get_info, because the table exists.
+-compile({inline, [assert_type_non_neg_integer/1]}).
+-spec assert_type_non_neg_integer(any()) -> non_neg_integer().
+assert_type_non_neg_integer(X) -> X.
 
 %% Cleanup
 -spec call_user_handle_down(server_pid(), state()) -> ok.
--- a/src/cets.erl
+++ b/src/cets.erl
@@ -667,14 +667,17 @@ handle_get_info(
     #{
         table => Tab,
         nodes => lists:usort(pids_to_nodes([self() | Servers])),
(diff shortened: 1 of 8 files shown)


src/cets.erl
```diff
@@ -601,12 +613,14 @@ handle_get_info(
...
         nodes => lists:usort(pids_to_nodes([self() | Servers])),
-        size => ets:info(Tab, size),
-        memory => ets:info(Tab, memory),
+        size => assert_integer(ets:info(Tab, size)),
+        memory => assert_integer(ets:info(Tab, memory)),
         ack_pid => AckPid,
         opts => Opts
     }.
 
+assert_integer(X) when is_integer(X) -> X.
```
erszcz:
```suggestion
-spec assert_integer(any()) -> integer().
assert_integer(X) when is_integer(X) -> X.
```

We have to add a spec to let Gradualizer now that after the guard succeeds the return value is actually an `integer()`. Otherwise, it would assume `-spec assert_integer(any()) -> any()` and therefore treat the return value as `any()` further on, which significantly relaxes any further checks against it.

arcusfelis: oh, interesting, dialyzer would still know that assert_integer returns an integer :)
But not gradualizer

```
Line 604 Column 2: Invalid type specification for function cets:handle_get_info/1. The success typing is (#{'ack_pid':=_, 'opts':=_, 'other_servers':=[pid()], 'tab':=atom() | ets:tid(), _=>_}) -> #{'ack_pid':=_, 'memory':=integer(), 'nodes':=[atom()], 'opts':=_, 'size':=integer(), 'table':=atom() | ets:tid()}
```

Will add

arcusfelis: ok, we don't really need to add assert, just spec is enough.

src/cets.erl
```diff
@@ -601,12 +613,17 @@ handle_get_info(
...
+        memory => returns_non_neg_integer(ets:info(Tab, memory)),
         ack_pid => AckPid,
         opts => Opts
     }.
 
+%% We are sure that the ETS table exists when running handle_get_info
+%% But gradualizer does not know that, so let us tell it with the spec
+-spec returns_non_neg_integer(any()) -> non_neg_integer().
+returns_non_neg_integer(X) -> X.
+
```
erszcz: If we don't want/need the guard, i.e. the dynamic check (which I think is better to have in general, unless we're 100% the result cannot be anything else), we can also do:

```erlang
-include_lib("gradualizer/include/gradualizer.hrl").

...
size => ?assert_type(ets:info(Tab, size), integer())
...
```

src/cets_call.erl
```diff
@@ -41,31 +41,41 @@ long_call(Server, Msg, Info) ->
...
 where(Pid) when is_pid(Pid) -> Pid;
-where(Name) when is_atom(Name) -> whereis(Name);
+where(Name) when is_atom(Name) -> pid_or_undefined(whereis(Name));
 where({global, Name}) -> global:whereis_name(Name);
-where({local, Name}) -> whereis(Name);
+where({local, Name}) -> pid_or_undefined(whereis(Name));
 where({via, Module, Name}) -> Module:whereis_name(Name).
 
+%% whereis/1 could return Port, so ignore Ports (makes Gradualizer happy)
+pid_or_undefined(undefined) -> undefined;
```
erszcz: Same story, if there's no spec then Gradualizer will assume `-spec pid_or_port(_) -> any()`.

src/cets.erl
arcusfelis: oh, actually runtime check
```
-spec assert_non_neg_integer(any()) -> non_neg_integer().
assert_non_neg_integer(X) when is_integer(X), X >= 0 -> X.
```

makes dialyzer unhappy

```
src/cets.erl
Line 620 Column 2: Type specification cets:assert_non_neg_integer(any()) -> non_neg_integer() is a supertype of the success typing: cets:assert_non_neg_integer(non_neg_integer()) -> non_neg_integer()
```
:)

arcusfelis: also, don't wanna add gradualizer as a build dep just for the header, to make compilation faster :D

erszcz: Yeah, we might create a `gradualizer_support` package at some point to address the compilation issue.

PR #12, opened 2023-08-01
Make CETS joining logic more reliable

arcusfelis:
This issue addresses MIM-1968.

The changes include:
- Tests for cets_join
- Step function to write tests for joining procedure
- Fixes for the logic - we do more validation before applying the changes
- We set join_ref when we join - all nodes should have the same join_ref after joining.
- We actually check join_ref before sending messages, check check_server in cets module.
- We don't catch errors for long calls anymore - it confuses result types. It also leads to nested `{error, {error, _}_}` when there is long call inside of a long operation 
- There is an issue in cover code reports, so we have to make a map outside of `LOG_ERROR` call and pass it into it. I don't think there would be any performance issues (unless someone disables all log messages), but still we could fix it once OTP fixes cover tool. For now I prefer good cover reports - they actually allow to find errors in logic https://github.com/erlang/otp/issues/7531

### DIFF
--- a/src/cets_join.erl
+++ b/src/cets_join.erl
@@ -8,7 +8,9 @@
 -type join_ref() :: reference().
 
 -type step() ::
-    before_check_fully_connected
+    join_start
+    | before_retry
+    | before_check_fully_connected
     | before_unpause
     | {before_send_dump, cets:server_pid()}.
 
@@ -65,6 +67,7 @@ join_loop(LockKey, Info, LocalPid, RemotePid, Start, JoinOpts) ->
     Retries = 1,
     case global:trans(LockRequest, F, Nodes, Retries) of
         aborted ->
+            run_step(before_retry, JoinOpts),
             ?LOG_ERROR(Info#{what => join_retry, reason => lock_aborted}),
             join_loop(LockKey, Info, LocalPid, RemotePid, Start, JoinOpts);
         Result ->
@@ -72,6 +75,7 @@ join_loop(LockKey, Info, LocalPid, RemotePid, Start, JoinOpts) ->
     end.
 
 join2(_Info, LocalPid, RemotePid, JoinOpts) ->
+    run_step(join_start, JoinOpts),
     JoinRef = make_ref(),
     %% Joining is a symmetrical operation here - both servers exchange information between each other.
     %% We still use LocalPid/RemotePid in names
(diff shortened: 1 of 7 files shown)


src/cets_join.erl
```diff
@@ -129,3 +179,72 @@ apply_resolver_for_sorted(
...
+            ?LOG_ERROR(Log),
+            error(check_same_join_ref_failed)
+    end.
+
+pid_to_join_ref(Pid) ->
+    #{join_ref := JoinRef} = cets:info(Pid),
+    JoinRef.
+
+-spec run_step(step(), join_opts()) -> ok.
+run_step(Step, #{step_handler := F}) ->
```
chrzaszcz: For me this is counter-intuitive, because this function isn't actually running the step, but runs some custom injected logic between steps. It's a lot of bending the code to fit the tests, and gives the impression of being overly complicated...

src/cets_join.erl
```diff
@@ -38,29 +71,35 @@ join_loop(LockKey, Info, LocalPid, RemotePid, Start) ->
...
             ?LOG_ERROR(Info#{what => join_retry, reason => lock_aborted}),
-            join_loop(LockKey, Info, LocalPid, RemotePid, Start);
+            join_loop(LockKey, Info, LocalPid, RemotePid, Start, JoinOpts);
         Result ->
             Result
     end.
 
-join2(_Info, LocalPid, RemotePid) ->
+join2(_Info, LocalPid, RemotePid, JoinOpts) ->
+    run_step(join_start, JoinOpts),
```
chrzaszcz: For me it would be cleaner if there was a `lists:foldl` over a list of functions performing the actual steps, and in tests you could use `meck` to inject some logic when needed - by changing the actual step function.

chrzaszcz (commented):
The code itself is OK for me 👍 
I only have an issue with the actual concept of steps, which are checkpoints between steps. I think a cleaner option would be to use a `fold` operation over a list of steps.

For example, see:
- https://github.com/esl/MongooseIM/blob/master/src/system_metrics/mongoose_system_metrics_collector.erl#L46
- https://github.com/esl/MongooseIM/blob/master/src/config/mongoose_config_parser_toml.erl#L142
- https://github.com/esl/MongooseIM/blob/master/src/graphql/mongoose_graphql_commands.erl#L70

I also consider the new syntax for logging errors a bit of verbose, and now we contruct the terms even if logs are disabled for the given level, which kind of defeats the purpose of the macro...

chrzaszcz (approved):
Looks cleaner with the checkpoints

PR #13, opened 2023-08-02
Use cets_ack process to collect replies

arcusfelis:
Addresses: MIM-1969

After adding leader_ops (i.e. insert_new), the logic in cets_call becomes a bit complicated.

Redesign inclides:
- use gen_server:call API instead of cets_call module.
- use a separate process to collect responses from nodes in cluster: cets_ack.
- use gen_server:reply once all replies collected (or the remote nodes crashed).

Motivation: 
- less code, less logic inside of the caller client process (i.e. it would only wait for a single gen_server reply now). 
- mon_tab ETS table could be removed, so no need for RPC calls to remove from it.

Initial benchmarking shows that the solution is pretty performant.

### DIFF
--- a/src/cets.erl
+++ b/src/cets.erl
@@ -482,16 +482,12 @@ set_other_servers(Servers, State = #{tab := Tab, ack_pid := AckPid}) ->
 pids_to_nodes(Pids) ->
     lists:map(fun node/1, Pids).
 
-ets_delete_keys(Tab, [Key | Keys]) ->
-    ets:delete(Tab, Key),
-    ets_delete_keys(Tab, Keys);
-ets_delete_keys(_Tab, []) ->
+ets_delete_keys(Tab, Keys) ->
+    [ets:delete(Tab, Key) || Key <- Keys],
     ok.
 
-ets_delete_objects(Tab, [Object | Objects]) ->
-    ets:delete_object(Tab, Object),
-    ets_delete_objects(Tab, Objects);
-ets_delete_objects(_Tab, []) ->
+ets_delete_objects(Tab, Objects) ->
+    [ets:delete_object(Tab, Object) || Object <- Objects],
     ok.
 
 %% Handle operation from a remote node
--- a/src/cets.erl
+++ b/src/cets.erl
@@ -84,8 +84,9 @@
 
 -include_lib("kernel/include/logger.hrl").
 
+-type server_pid() :: pid().
 -type server_ref() ::
-    pid()
+    server_pid()
     | atom()
     | {local, atom()}
     | {global, term()}
@@ -101,16 +102,16 @@
     | {delete_objects, [term()]}
     | {insert_new, tuple()}
     | {leader_op, op()}.
--type remote_op() :: {remote_op, Op :: op(), From :: from(), AckPid :: pid()}.
+-type remote_op() :: {remote_op, Op :: op(), From :: from(), AckPid :: cets_ack:ack_pid()}.
 -type backlog_entry() :: {op(), from()}.
 -type table_name() :: atom().
 -type pause_monitor() :: reference().
 -type state() :: #{
     tab := table_name(),
-    ack_pid := pid(),
+    ack_pid := cets_ack:ack_pid(),
     %% Updated by set_other_servers/2 function only
-    other_servers := [pid()],
-    leader := pid(),
+    other_servers := [server_pid()],
+    leader := server_pid(),
     is_leader := boolean(),
     opts := start_opts(),
     backlog := [backlog_entry()],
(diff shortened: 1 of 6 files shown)


src/cets.erl
```diff
@@ -545,99 +522,94 @@ do_table_op({insert_new, Rec}, Tab) ->
...
+
+-spec replicate(op(), from(), state()) -> ok.
+replicate(_Op, From, #{other_servers := []}) ->
+    %% Skip replication
+    gen_server:reply(From, ok);
+replicate(Op, From, #{ack_pid := AckPid, other_servers := Servers}) ->
+    cets_ack:add(AckPid, From),
+    RemoteOp = {remote_op, Op, From, AckPid},
+    [send_remote_op(Server, RemoteOp) || Server <- Servers],
+    %% AckPid would call gen_server:reply(From, ok) ones all the remote servers reply
```
```suggestion
    %% AckPid would call gen_server:reply(From, ok) once all the remote servers reply
```

src/cets_ack.erl
```diff
@@ -0,0 +1,151 @@
+%% CETS Ack Helper Process
+%%
+%% Contains a list of processes, that are waiting for writes to finish.
+%% Collects acks from nodes in the cluster.
+%% When one of the remote nodes go down, the server stops waiting for acks from it.
```
chrzaszcz:
```suggestion
%% When one of the remote nodes goes down, the server stops waiting for acks from it.
```

src/cets_ack.erl
```diff
@@ -0,0 +1,151 @@
...
+    Name = list_to_atom(atom_to_list(Tab) ++ "_ack"),
+    gen_server:start_link({local, Name}, ?MODULE, [], []).
+
+%% Sets a list of servers to be used for newly added tasks
+-spec set_servers(ack_pid(), [server_pid()]) -> ok.
+set_servers(AckPid, Servers) ->
+    gen_server:cast(AckPid, {set_servers, Servers}),
+    ok.
+
+%% Adds a new task to wait replies
```
chrzaszcz:
```suggestion
%% Adds a new task to wait for replies
```

src/cets_ack.erl
```diff
@@ -0,0 +1,151 @@
...
+-spec add(ack_pid(), from()) -> ok.
+add(AckPid, From) ->
+    AckPid ! {add, From},
+    ok.
+
+%% Called by a remote server after an operation is applied.
+-spec ack(ack_pid(), from(), server_pid()) -> ok.
+ack(AckPid, From, RemotePid) ->
+    %% nosuspend makes message sending unreliable
```
chrzaszcz: This looks confusing, because there is no `nosuspend`. Did you intend to say this?
```suggestion
    %% nosuspend is not used, because it would make message sending unreliable
```

test/cets_SUITE.erl
```diff
@@ -609,10 +670,73 @@ insert_into_keypos_table(_Config) ->
...
+    {ok, T} = cets:table_name(Pid),
+    #{table := T} = cets:info(Pid).
+
 info_contains_opts(_Config) ->
     {ok, Pid} = cets:start(info_contains_opts, #{type => bag}),
     #{opts := #{type := bag}} = cets:info(Pid).
 
+%% Cases to improve code coverage
+
+unknown_down_message_is_ignored(_Config) ->
```
NelsonVides, line 686: Would be good to get these tests into MIM where we use those `UNEXPECTED_CALL`/`UNEXPECTED_CAST`/`UNEXPECTED_INFO` macros that are entirely untested 😄

chrzaszcz (commented):
Looks good, it is indeed more readable and uses the OTP functionality better. I added minor comments.

src/cets_ack.erl
arcusfelis: yeah, it is for the future if someone tried to add nosuspend there again.

noconnect is probably fine, because if there is no connection to the node - the servers would receive `DOWN` message from an unreachable server and ack would not be needed to be sent.

nosuspend is usually used when the code is checking that nosuspend is returned and spawns a new process to send a message. But I think we probably would set new async_dist=true flag for the server process, instead of nosuspend option (though in a separate PR).

src/cets_ack.erl
```diff
@@ -0,0 +1,151 @@
...
+    case State of
+        #{From := Servers} ->
+            handle_updated(From, RemotePid, Servers, State);
+        _ ->
+            %% Ignore unknown From
+            State
+    end.
+
+-spec handle_updated(from(), server_pid(), [server_pid(), ...], state()) -> state().
+handle_updated(From, RemotePid, Servers, State) ->
```
NelsonVides, line 139: Very smart idea, I like it :)

NelsonVides (approved):
Looks great to me 😄

PR #14, opened 2023-08-04
Use makefile

arcusfelis:
Changes:
- Use makefile for commands - we could run `rebar3 command` manually, but it is useful to have a command which runs all the checks.
- Put plugins into rebar3 profiles - faster CI (no need to compile fmt plugin when running gradualizer task).
- Enable cover for tests: we want to see if we have any dead code in tests, because if we add a case, but forget to add it into `all/0` list - it would never be called.
- Made a separate step for `pre_test` - we don't need it, but CI logs look better with it (we can ignore compilation output there).

WIP, merge gradualizer PR first.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -18,12 +18,8 @@ jobs:
 
     steps:
     - uses: actions/checkout@v3
-    - name: Compile
-      run: rebar3 compile
     - name: Run tests
-      run: rebar3 as test ct --sname=ct1 --cover
-    - name: Send test coverage report
-      run: rebar3 as test codecov analyze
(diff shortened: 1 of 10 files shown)


NelsonVides:
If all you want is to run several rebar3 commands as a single `make X`, you can use aliases in rebar3. See for example `worker_pool` trick here: https://github.com/inaka/worker_pool/blob/d11750321cade5b4aa516447d9063453a159a7eb/rebar.config#L73 I'd prefer that honestly, one less file and config to worry about even if that is just a makefile.

arcusfelis:
@NelsonVides could be aliases :)

PR #16, opened 2023-08-04
Format with AdRoll/rebar3_format

arcusfelis:
WIP do not merge.
Tried https://github.com/AdRoll/rebar3_format
(it has a lot of config options though - could be useful to match MIM style).


arcusfelis:
Could close, it is PR for demo only.

PR #18, opened 2023-08-04
Parallel tests

arcusfelis:
Changes:
- parallel tests.
- stop CETS servers on exit.
- repeat 3 times.
- use helpers to name tables.
- use helpers to spawn common configurations (i.e. 2 joined nodes).

- NO new tests.
- NO modification to the main code.

Motivation:
- Naming servers is hard :)
- Code looked different in each testcase (because of manual naming)

### DIFF
--- a/test/cets_SUITE.erl
+++ b/test/cets_SUITE.erl
@@ -106,66 +106,61 @@ init_per_testcase_generic(Name, Config) ->
 end_per_testcase(_, _Config) ->
     ok.
 
-inserted_records_could_be_read_back(_Config) ->
-    start_local(ins1, #{}),
-    cets:insert(ins1, {alice, 32}),
-    [{alice, 32}] = ets:lookup(ins1, alice).
+inserted_records_could_be_read_back(Config) ->
+    Tab = make_name(Config),
+    start_local(Tab),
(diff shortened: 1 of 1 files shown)


NelsonVides (commented):
Looks good, for a moment I thought well this is not any faster, CI takes the same time, but right, tests are ran three times now and it still takes the same time, so that's a good speedup. Anyway, why do we want to run the tests three times?

arcusfelis:
> Anyway, why do we want to run the tests three times?

@NelsonVides We could run them two times.
The main reason, I don't wanna break support for rerun-able tests in the future :)
I.e. I can be sure that I can set manually `{repeat_until_any_fail, 1000}` and go drink coffee, if I need to debug something flaky.

But tests are so fast, that there is no visible difference between 2 times and 3 :D 

```
time ./run_test.sh
===> Verifying dependencies...
===> Analyzing applications...
===> Compiling cets
===> Running Common Test suites...
%%% cets_SUITE: ........................................................................................................................................................................................................................
All 216 tests passed.
./run_test.sh  3.80s user 1.34s system 115% cpu 4.466 total
```

NelsonVides (approved):
LGTM

PR #19, opened 2023-08-05
Add Hank (dead code finder)

arcusfelis:
Looks useful to remove unused arguments and stuff.

### DIFF
--- a/rebar.config
+++ b/rebar.config
@@ -21,11 +21,15 @@
     {cover_tests, [
         {ct, "--sname=ct1 --cover"}
     ]},
+    {find_dead_code, [
+        {do, "default as hank hank"}
+    ]},
     {all, [
         format,
         run_tests,
         xref,
-        dialyzer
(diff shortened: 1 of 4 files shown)


rebar.config
```diff
@@ -21,21 +21,28 @@
...
         {plugins, [erlfmt]}
     ]},
+    {hank, [
+        {plugins, [rebar3_hank]}
+    ]},
     {test, [
         {plugins, [
-            {rebar3_codecov, "0.4.0"}
+            {rebar3_codecov, "0.5.0"}
         ]}
```
NelsonVides, line 46: Do we need so many profiles? Why can't all of them be in the test profile together? 🤔

arcusfelis: Because rebar does not do lazy loading for plugins sadly :( 
There is a big chance I could just move all of them into `project_plugins` and don't care about extra 10 seconds per task on CI :D  (later probably)

PR #20, opened 2023-08-05
Use rebar3 aliases

arcusfelis:
I've tried to use makefiles to automate running 3 commands I always do, but @NelsonVides suggested to use rebar3 aliases.

- Added `rebar3 all` that runs all the checks before making a commit or PR.
- Move `rebar3 fmt` into a profile, because it takes a bit of time to build formatting tool and its deps (i.e. faster CI for tasks like dialyzer and build).
- Remove my favourite `run_test.sh`

PS: syntax for commands with aliases is weird (i.e. do...default).

### DIFF
--- a/rebar.config
+++ b/rebar.config
@@ -24,6 +24,7 @@
     {all, [
         format,
         run_tests,
+        xref,
         dialyzer
     ]}
 ]}.
--- a/rebar.config
+++ b/rebar.config
@@ -8,10 +8,30 @@
     ]}
 ]}.
 
-%% Enables "rebar3 fmt" command
-{project_plugins, [erlfmt]}.
+{alias, [
+    {format, [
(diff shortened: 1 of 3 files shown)


rebar.config
```diff
@@ -8,10 +8,31 @@
...
+        {ct, "--sname=ct1"}
+    ]},
+    {cover_tests, [
+        {ct, "--sname=ct1 --cover"}
+    ]},
+    {all, [
+        format,
+        run_tests,
+        xref,
+        dialyzer
```
NelsonVides, line 28: You might want to add here hank or gradualizer or any other tool you might add later :)

NelsonVides (approved):

rebar.config
arcusfelis: yes, right

PR #21, opened 2023-08-11
More dialyzer checks

arcusfelis:
MIM-1981.

Changes:
- adds extra_return/missing_return checks
- fix specs (uses the OTP specs)


NelsonVides (approved):
LGTM

PR #22, opened 2023-08-11
Tweak get_leader to return same error

arcusfelis:
get_leader could return noproc or ets table not found (i.e. metadata table).
Tweak code to return only noproc error.


NelsonVides (approved):
Fine, consistency 👌🏽

PR #23, opened 2023-08-22
Fix logger cover

arcusfelis:
The PR addresses MIM-2010:
- Multiline logging calls report one line as not covered by tests. See diff for the example of such calls.

Changes:
- Undo the changes introduced https://github.com/esl/cets/pull/12 - i.e. `Log = #{...}, ?LOG_ERROR(Log)` workaround.
- Run tests with logging disabled instead (additionally to the regular run).

More info about what it does https://github.com/erlang/otp/issues/7531


NelsonVides (approved):
Well, if we have to then be it, good to find out 🤷🏽‍♂️

PR #24, opened 2023-08-23
Improved discovery backend

arcusfelis:
Adresses MIM-2018.

Make cets_discovery module more async and better tested.
This should avoid long locking time at startup in some cases (for example, there are a lot of dead nodes returned by the discovery backend).

Changes:
- Make discovery gen_server non-blocking.
- Adding wait_for_ready functionality - could be useful to avoid not-fully ready servers to be exposed to the users (i.e. to avoid accepting incoming connection on a new node, before it copied the session table, for example)
- Smarter timeouts for retries. We use a different timeout after an error. We also don't call get_nodes too often once started.
- Specs.
- More tests with better coverage.
- Fixes small race conditions in insert_new tests (rare, but they were found by rerunning tests).
- Use peer module to do RPCs.
- Some netsplit tests.

You can find MIM branch that uses this code here: https://github.com/esl/MongooseIM/pull/4111 (for testing)

Command to test helm (ensure you use k8s from Docker Desktop and having helm installed):
```bash
# Setup DB using MIM test scripts:
DB="pgsql" ./tools/setup-db.sh
# Ensure MIM stopped...
killall beam.smp
# Get helm scripts
git clone https://github.com/esl/MongooseHelm.git
cd MongooseHelm
# Ensure old hem cluster is dead
helm uninstall mim-test
# Test with these changes
helm install mim-test MongooseIM --set replicaCount=3 --set image.tag=PR-4111 --set persistentDatabase=rdbms --set rdbms.username=ejabberd --set rdbms.password=mongooseim_secret --set rdbms.database=ejabberd --set volatileDatabase=cets
# Check that all is fine in logs:
kubectl get pod
kubectl logs mongooseim-0
kubectl exec -it mongooseim-0 -- mongooseimctl cets systemInfo
```

### DIFF
--- a/src/cets_discovery.erl
+++ b/src/cets_discovery.erl
@@ -317,10 +317,10 @@ choose_retry_type(_) ->
 %% Returns timeout in milliseconds to retry calling the get_nodes function.
 %% get_nodes is called after add_table without waiting.
 %% It is also would be retried without waiting if should_retry_get_nodes set to true.
--spec retry_type_to_timeout(retry_type()) -> 5000 | 1000 | 180000.
-retry_type_to_timeout(initial) -> 5000;
-retry_type_to_timeout(after_error) -> 1000;
-retry_type_to_timeout(regular) -> 180000.
+-spec retry_type_to_timeout(retry_type()) -> non_neg_integer().
+retry_type_to_timeout(initial) -> timer:seconds(5);
+retry_type_to_timeout(after_error) -> timer:seconds(1);
+retry_type_to_timeout(regular) -> timer:minutes(5).
 
 -spec cancel_old_timer(state()) -> ok.
 cancel_old_timer(#{timer_ref := OldRef}) when is_reference(OldRef) ->
--- a/src/cets_discovery.erl
+++ b/src/cets_discovery.erl
@@ -179,6 +179,8 @@ handle_info({nodeup, Node}, State) ->
     State2 = remove_node_from_unavailable_list(Node, State),
     {noreply, try_joining(State2)};
 handle_info({nodedown, _Node}, State) ->
+    %% Do another check to update unavailable_nodes list
+    self() ! check,
     {noreply, State};
 handle_info({joining_finished, Results}, State) ->
     {noreply, handle_joining_finished(Results, State)};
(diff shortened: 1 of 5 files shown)


src/cets_join.erl
```diff
@@ -201,8 +201,8 @@ check_do_not_overlap(LocPids, RemPids) ->
         Overlap ->
             ?LOG_ERROR(#{
                 what => check_do_not_overlap_failed,
-                local_servers => LocPids,
-                remote_servers => RemPids,
+                local_servers => LocSet,
+                remote_servers => RemSet,
```
NelsonVides: Won't these be more unreadable in the logs than a list? I'd say it depends on whether the list had duplicates. If ordsets are simply ordering the list into a data structure that is faster to read or calculate intersections then I'd prefer we print the list, it's just more readable 🤔

src/cets_discovery.erl
```diff
@@ -120,24 +211,118 @@ code_change(_OldVsn, State, _Extra) ->
...
+%% - a list of nodes is received from the discovery backend
+-spec try_joining(state()) -> state().
+try_joining(State = #{join_status := running}) ->
+    State#{should_retry_join := true};
+try_joining(State = #{join_status := not_running, nodes := Nodes, tables := Tables}) ->
+    Self = self(),
+    AvailableNodes = nodes(),
+    spawn_link(fun() ->
+        %% We only care about connected nodes here
+        %% We do not wanna try to connect here - we do it in ping_not_connected_nodes/1
```
NelsonVides:
```suggestion
        %% We do not want to try to connect here - we do it in ping_not_connected_nodes/1
```

src/cets_discovery.erl
```diff
@@ -120,24 +211,118 @@ code_change(_OldVsn, State, _Extra) ->
...
+choose_retry_type(_) ->
+    regular.
+
+%% Returns timeout in milliseconds to retry calling the get_nodes function.
+%% get_nodes is called after add_table without waiting.
+%% It is also would be retried without waiting if should_retry_get_nodes set to true.
+-spec retry_type_to_timeout(retry_type()) -> 5000 | 1000 | 180000.
+retry_type_to_timeout(initial) -> 5000;
+retry_type_to_timeout(after_error) -> 1000;
+retry_type_to_timeout(regular) -> 180000.
```
NelsonVides: I don't know from the top of my mind how many minutes `180000`ms is, so maybe at a comment, `180000ms = 5min`, or, call `timer:minutes(5)`, to make it more readable (comment and direct value might be "more performant though" xD)

NelsonVides (commented):
Just a few very small comments

src/cets_join.erl
arcusfelis: oh yea, I did the change for logs to make logs easier to read.
get_pids() does return `[Server | ordeset()]`. So, I can just force to use ordsets everywhere, no need to allow non-sorted lists.

NelsonVides (approved):
LGTM 🔥

PR #29, opened 2023-08-31
Check that nodes are fully connected before allowing to join

arcusfelis:
Checks that nodes are fully connected (could send messages to each other) before allowing to join. MIM-2023

It should allow to not join with a node, which is netsplitted or have a wrong network configuration (and prevent_overlapping_partitions is false).

It **does not** address case when a node permanently looses connectivity to some part of the cluster. We would have to do it separately, because there are many ways to choose which nodes should be disconnected when they loose partial connectivity to the cluster (i.e. we would have to decide on voting + healthcheck)

Also, the motivation is to be able to not allow bad (not fully connected in the sense of erlang distribution) nodes to rejoin, after we would design code which would kick nodes out of the cluster.

### DIFF
--- a/rebar.config
+++ b/rebar.config
@@ -36,6 +36,9 @@
         {plugins, [erlfmt]}
     ]},
     {test, [
+        {deps, [
+            {logger_debug_h, "0.1.0"}
+        ]},
         {plugins, [
             {rebar3_codecov, "0.6.0"}
         ]}
--- a/test/cets_SUITE.erl
+++ b/test/cets_SUITE.erl
@@ -109,6 +109,7 @@ cases() ->
         code_change_returns_ok_for_ack,
         run_spawn_forwards_errors,
         run_tracked_failed,
+        run_tracked_logged,
         long_call_to_unknown_name_throws_pid_not_found,
         send_leader_op_throws_noproc
     ].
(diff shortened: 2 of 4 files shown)


rebar.config
```diff
@@ -36,6 +36,9 @@
         {plugins, [erlfmt]}
     ]},
     {test, [
+        {deps, [
+            {logger_debug_h, "0.1.0"}
```
NelsonVides, line 40: Maaaan, that's as useful as it is lazy xD

test/cets_SUITE.erl
```diff
@@ -1170,6 +1181,15 @@ info_contains_opts(Config) ->
...
 
+check_could_reach_each_other_fails(_Config) ->
+    matched =
+        try
+            cets_join:check_could_reach_each_other([self()], [bad_node_pid()])
+        catch
+            error:check_could_reach_each_other_failed ->
+                matched
+        end.
+
```
NelsonVides: It'd rather do something like
```suggestion
        try
            cets_join:check_could_reach_each_other([self()], [bad_node_pid()])
            ct:fail(...)
        catch
            error:check_could_reach_each_other_failed ->
                ok
        end.

```

test/cets_SUITE.erl
```diff
@@ -1315,6 +1362,62 @@ disco_connects_to_unconnected_node(Config) ->
     cets_discovery:add_table(Disco, Tab),
     ok = cets_discovery:wait_for_ready(Disco, 5000).
 
+%% Joins from a bad (not fully connected) node
+%% Join process should check if nodes could contact each other before allowing to join
+joining_not_fully_connected_node_is_not_allowed(Config) ->
+    #{ct3 := Node3, ct5 := Node2} = proplists:get_value(peers, Config),
```
NelsonVides: Why not calling ct5 := Node5? Just to keep numbers matching the node number? Maybe it's easier that way to keep in mind the configuration of such node?

arcusfelis: Pid1, Pid5... hard to read test.
Given 2 nodes, we do something on first node Node1 and do something on the second node Node5 (surprise).

rebar.config
arcusfelis: Yea, it is useful to check that we actually log something :) 
Made it as a dep, so could use for MIM tests if I have to.

test/cets_SUITE.erl
arcusfelis: rewrote with eunit macros

test/cets_SUITE.erl
arcusfelis: ok, Pid5/Node5/Peer5. Also separated Nodes from Peers in names.

test/cets_SUITE.erl
NelsonVides: Soooo much better 😃

NelsonVides (approved):

PR #32, opened 2023-09-11
Add cets_status module

arcusfelis:
Add cets_status module

The code gets a list of nodes, gets a list of servers and compares them.
After that it produces a map like:

```erlang
-type info() :: #{
    %% Nodes that are connected to us and have the CETS disco process started.
    available_nodes := [node()],
    %% Nodes that do not respond to our pings.
    unavailable_nodes := [node()],
    %% Nodes with stopped disco
    remote_nodes_without_disco := [node()],
    %% Nodes that has our local tables running (but could also have some unknown tables).
    %% All joined nodes replicate data between each other.
    joined_nodes := [node()],
    %% Nodes that have more tables registered than the local node.
    remote_nodes_with_unknown_tables := [node()],
    remote_unknown_tables := [table_name()],
    %% Nodes that are available, but do not host one of our local tables.
    remote_nodes_with_missing_tables => [node()],
    remote_missing_tables := [table_name()],
    %% Nodes that replicate at least one of our local tables to a different list of nodes
    %% (could temporary happen during a netsplit)
    conflict_nodes := [node()],
    conflict_tables := [table_name()],
    %% Nodes that are extracted from the discovery backend.
    discovered_nodes := [node()],
    %% True, if discovery backend returned the list of nodes the last time we've tried
    %% to call it.
    discovery_works := boolean()
}.
```

The map could be used for getting the status of the cluster.

This functionality is used in MongooseIM like that https://github.com/esl/MongooseIM/pull/4116

### DIFF
--- a/src/cets_status.erl
+++ b/src/cets_status.erl
@@ -8,83 +8,142 @@
 -type disco_name() :: atom().
 
 -type info() :: #{
-    unavailable_nodes => [node()],
-    available_nodes => [node()],
-    joined_nodes => [node()],
-    partially_joined_nodes => [node()],
-    partially_joined_tables => [table_name()],
-    discovered_nodes => [node()],
-    discovery_works => boolean()
+    %% Nodes that are connected to us and have the CETS disco process started.
+    available_nodes := [node()],
+    %% Nodes that do not respond to our pings.
+    unavailable_nodes := [node()],
+    %% Nodes with stopped disco
+    remote_nodes_without_disco := [node()],
+    %% Nodes that has our local tables running (but could also have some unknown tables).
+    %% All joined nodes replicate data between each other.
+    joined_nodes := [node()],
+    %% Nodes that have more tables registered than the local node.
+    remote_nodes_with_unknown_tables := [node()],
+    remote_unknown_tables := [table_name()],
+    %% Nodes that are available, but do not host one of our local tables.
+    remote_nodes_with_missing_tables => [node()],
+    remote_missing_tables := [table_name()],
+    %% Nodes that replicate at least one of our local tables to a different list of nodes
+    %% (could temporary happen during a netsplit)
+    conflict_nodes := [node()],
+    conflict_tables := [table_name()],
+    %% Nodes that are extracted from the discovery backend.
+    discovered_nodes := [node()],
+    %% True, if discovery backend returned the list of nodes the last time we've tried
+    %% to call it.
+    discovery_works := boolean()
 }.
 
 -spec status(disco_name()) -> info().
 status(Disco) when is_atom(Disco) ->
     %% The node lists could not match for different nodes
     %% because they are updated periodically
-    #{unavailable_nodes := UnNodes, nodes := Nodes, tables := Tables} =
+    #{unavailable_nodes := UnNodes, nodes := DiscoNodes, tables := Tables} =
         Info = cets_discovery:system_info(Disco),
-    NodesSorted = lists:sort(Nodes),
-    AvailNodes = available_nodes(Disco),
-    {JoinedNodes, PartTables} = filter_joined_nodes(AvailNodes, Tables, Disco),
-    PartNodes = AvailNodes -- JoinedNodes,
+    DiscoNodesSorted = lists:sort(DiscoNodes),
+    OnlineNodes = [node() | nodes()],
+    AvailNodes = available_nodes(Disco, OnlineNodes),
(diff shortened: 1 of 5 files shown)


src/cets_status.erl
```diff
@@ -0,0 +1,214 @@
...
+%% Nodes, that host the discovery process
+-spec available_nodes(disco_name(), [node(), ...]) -> [node()].
+available_nodes(Disco, OnlineNodes) ->
+    [Node || Node <- OnlineNodes, is_disco_running_on(Node, Disco)].
+
+remote_nodes_without_disco(DiscoNodes, AvailNodes, OnlineNodes) ->
+    [
+        Node
+     || Node <- DiscoNodes, lists:member(Node, OnlineNodes), not lists:member(Node, AvailNodes)
+    ].
```
NelsonVides: Formatting here is weird, maybe something like below or so
```suggestion
remote_nodes_without_disco(DiscoNodes, AvailNodes, OnlineNodes) ->
    [ Node || Node <- DiscoNodes,
              lists:member(Node, OnlineNodes),
              not lists:member(Node, AvailNodes)
    ].
```

src/cets_status.erl
```diff
@@ -0,0 +1,214 @@
...
+
+-spec is_disco_running_on(node(), disco_name()) -> boolean().
+is_disco_running_on(Node, Disco) ->
+    is_pid(rpc:call(Node, erlang, whereis, [Disco])).
+
+-spec gather_tables_and_replication_nodes(
+    AvailNodes :: [node()], Disco :: disco_name()
+) ->
+    OtherTabNodes :: node_to_tab_nodes_map().
```
NelsonVides: This is also super strange formatting, I really dislike breaking on the final parenthesis, at least something like this would be better
```suggestion
-spec gather_tables_and_replication_nodes(AvailNodes :: [node()], Disco :: disco_name()) ->
    OtherTabNodes :: node_to_tab_nodes_map().
```

src/cets_status.erl
```diff
@@ -0,0 +1,214 @@
...
+    get_table_to_other_nodes_map(Node, Tables).
+
+%% Returns nodes for each table in the Tables list
+-spec get_table_to_other_nodes_map(node(), [table_name()]) -> tab_nodes_map().
+get_table_to_other_nodes_map(Node, Tables) ->
+    maps:from_list([{Table, get_node_list_for_table(Node, Table)} || Table <- Tables]).
+
+-spec get_tables_list_on_node(Node :: node(), Disco :: disco_name()) -> [table_name()].
+get_tables_list_on_node(Node, Disco) ->
+    case rpc:call(Node, cets_discovery, get_tables, [Disco]) of
```
NelsonVides: Just a general question, why not using the newer `erpc` module directly?

src/cets_status.erl
```diff
@@ -0,0 +1,214 @@
...
+        {ok, Tables} ->
+            Tables;
+        _ ->
+            []
+    end.
+
+-spec get_node_list_for_table(Node :: node(), Table :: table_name()) ->
+    Nodes :: ordsets:ordset(node()).
+get_node_list_for_table(Node, Table) ->
+    case catch rpc:call(Node, cets, other_nodes, [Table]) of
```
NelsonVides: Like, to better distinguish these places where you need to do `catch`, `erpc` seems to have a cleaner interface for that.

NelsonVides (commented):
The set operations happening here over maps and lists and proplists and maps of proplists and lists of maps and maps of maps of lists and just about every combination possible, is actually quite entertaining :D

Just a few minor comments.

src/cets_status.erl
arcusfelis: formatting here is autoformatted, so

src/cets_status.erl
arcusfelis: oh, actually not xdd
we don't need catch here. But we don't need erpc here too probably.

src/cets_status.erl
arcusfelis: I am too lazy to rewrite stuff in parallel fashion here, maybe it is time :)

src/cets_status.erl
arcusfelis: Could just rewrite.
The code is autoformatted.

src/cets_status.erl
arcusfelis: It would not work on OTP 24 though, only 25 :)

NelsonVides (approved):
LGTM

PR #33, opened 2023-10-02
Add cets:insert_serial/2

arcusfelis:
All `insert_serial' calls are sent to the leader node first.

Similar to `insert_new/2`, but overwrites the data silently on conflict.
 It could be used to update entries, which use not node-specific keys.
 
 This function ensures that all nodes would agree on the value, if they try to insert (OR overwrite) a record with the same key at once.
 
 It is used by https://github.com/esl/MongooseIM/pull/4136

### DIFF
--- a/src/cets.erl
+++ b/src/cets.erl
@@ -23,6 +23,7 @@
     insert/2,
     insert_many/2,
     insert_new/2,
+    insert_serial/2,
     delete/2,
     delete_many/2,
     delete_object/2,
@@ -63,6 +64,7 @@
     insert/2,
     insert_many/2,
     insert_new/2,
+    insert_serial/2,
     delete/2,
     delete_many/2,
     delete_object/2,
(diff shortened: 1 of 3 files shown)


arcusfelis:
@chrzaszcz Propose better name for insert_serial, if you want. insert_serialized? upsert_though_the_main_node? :) 
**insert_leader**? leader_insert? insert_at_leader? insert_though_leader?

src/cets.erl
```diff
@@ -243,12 +245,22 @@ insert_many(Server, Records) when is_list(Records) ->
...
 
+%% @doc Serialized version of `insert/2'.
+%%
+%% All `insert_serial' calls are sent to the leader node first.
+%%
+%% Similar to `insert_new/2', but overwrites the data silently on conflict.
+%% It could be used to update entries, which use not node-specific keys.
+-spec insert_serial(server_ref(), tuple()) -> ok.
+insert_serial(Server, Rec) when is_tuple(Rec) ->
+    ok = cets_call:send_leader_op(Server, {insert, Rec}).
```
chrzaszcz, line 262: So we send the operation to the leader node, but we skip the `leader_op` check. Why?

test/cets_SUITE.erl
```diff
@@ -381,6 +383,19 @@ insert_new_fails_if_the_local_server_is_dead(_Config) ->
         exit:{noproc, {gen_server, call, _}} -> ok
     end.
 
+insert_serial_works(Config) ->
+    #{pid1 := Pid1, tab1 := Tab1, tab2 := Tab2} = given_two_joined_tables(Config),
+    ok = cets:insert_serial(Pid1, {a, 1}),
+    [{a, 1}] = cets:dump(Tab1),
+    [{a, 1}] = cets:dump(Tab2).
+
+insert_serial_overwrites_data(Config) ->
```
chrzaszcz, line 397: I don't see the point of this test, because if I replace `insert_serial` with just `insert`, the test would also pass. IMO it should check that `insert_serial` is in some way different than just `insert`.

chrzaszcz (commented):
I added some comments.

src/cets.erl
arcusfelis: good catch. Generally, I wanted to avoid match for true | false, returned by insert_new. But I've also skipped is_leader check, added a test and a fix.

test/cets_SUITE.erl
arcusfelis: good, added three tests to show difference between insert_new, insert and insert_serial.

insert_serial_overwrites_data is still a very basic testcase, so we could keep it (I remember first CETS review asked to add small testcases :)).

src/cets.erl
arcusfelis: also, apparently, insert also returns true (instead of ok). So, I don't need to match for ok in the return before replicating in handle_leader_op. Ah, prolog api :)

src/cets_call.erl
```diff
@@ -70,7 +70,7 @@ send_leader_op(Server, Op) ->
 
 send_leader_op(Server, Op, Backoff) ->
     Leader = cets:get_leader(Server),
-    Res = maybe_sync_operation(Leader, {leader_op, Op}),
```
chrzaszcz: So maybe we don't need this change?

arcusfelis: Could revert that one too.

chrzaszcz (approved):
👌

PR #34, opened 2023-10-05
Do not log check_do_not_overlap_failed error, if the cluster is already joined

arcusfelis:
Do not log check_do_not_overlap_failed error, if the cluster is already joined MIM-2061.

Two servers could discover each other at the same time and try to join. One server would join, while another will wait for a lock, get a lock and print the error in the old version.

In the new version we actually check if it is an overlap, or just fully joined cluster. In case it is an overlap we still print an error and retry. In case pids on both remote and local sides are the same, we could just do nothing (i.e. the cluster looks properly connected at this point).

### DIFF
--- a/test/cets_SUITE.erl
+++ b/test/cets_SUITE.erl
@@ -143,7 +143,7 @@ only_for_logger_cases() ->
     [
         run_tracked_logged_check_logger,
         logs_are_printed_when_join_fails_because_servers_overlap,
-        join_done_already_while_waiting_for_lock_so_do_nothing_silently
+        join_done_already_while_waiting_for_lock_so_do_nothing
     ].
 
 seq_cases() ->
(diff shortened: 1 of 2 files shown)


test/cets_SUITE.erl
```diff
@@ -141,7 +141,9 @@ cases() ->
 
 only_for_logger_cases() ->
     [
-        run_tracked_logged_check_logger
+        run_tracked_logged_check_logger,
+        logs_are_printed_when_join_fails_because_servers_overlap,
+        join_done_already_while_waiting_for_lock_so_do_nothing_silently
```
chrzaszcz: pleonasm
```suggestion
        join_done_already_while_waiting_for_lock_so_do_nothing
```

test/cets_SUITE.erl
```diff
@@ -696,6 +718,65 @@ join_retried_if_lock_is_busy(Config) ->
...
+            ok
+    end,
+    F = fun(_) -> ok end,
+    %% It is to just match logs
+    LogRef = make_ref(),
+    Info = #{log_ref => LogRef},
+    %% Get the lock in a separate process
+    spawn_link(fun() ->
+        ok = cets_join:join(Lock, Info, Pid1, Pid3, #{checkpoint_handler => SleepyF}),
+        Me ! second_join_returns
```
chrzaszcz: Same as in line 759? Shouldn't it be 'first join'?

test/cets_SUITE.erl
```diff
@@ -696,6 +718,65 @@ join_retried_if_lock_is_busy(Config) ->
...
+    logger_debug_h:start(#{id => ?FUNCTION_NAME}),
+    Me = self(),
+    {ok, Pid1} = start_local(make_name(Config, 1)),
+    {ok, Pid2} = start_local(make_name(Config, 2)),
+    {ok, Pid3} = start_local(make_name(Config, 3)),
+    {ok, Pid4} = start_local(make_name(Config, 4)),
+    Lock = lock_name(Config),
+    ok = cets_join:join(Lock, #{}, Pid1, Pid2, #{}),
+    ok = cets_join:join(Lock, #{}, Pid3, Pid4, #{}),
+    SleepyF = fun
```
chrzaszcz: Maybe extract to a helper function? The test case is getting long.

chrzaszcz (commented):
Looks good in general. I added some comments.

chrzaszcz (approved):
Looks ok

PR #35, opened 2023-10-05
Ensure that unavailable_nodes list is a subset of discovered_nodes

arcusfelis:
Addresses "MIM-2054 CETS status: removed nodes from disco should not be listed in a list of unavailable nodes"

Changes:
- Ensure that unavailable_nodes list is a subset of discovered_nodes
- Add status_discovered_nodes testcase (a test  for this field was missing)
- Use ordsets in types for nodes (because they are ordsets)
- Define `DiscoOpts` in tests instead of constructing the same map twice (small refactoring)

Nodes could be removed from the discovery_nodes table after some time.
After that point, the removed node should not appear in the systemInfo API.
But we currently we can have:

```json
./_build/mim1/rel/mongooseim/bin/mongooseimctl cets systemInfo
{
  "data" : {
    "cets" : {
      "systemInfo" : {
        "unavailableNodes" : [
          "badnode@localhost"
        ],
        "discoveredNodes" : [
          "mongooseim2@localhost",
          "mongooseim3@localhost",
          "mongooseim@localhost"
        ],
        "availableNodes" : [
          "mongooseim@localhost",
          "mongooseim2@localhost",
          "mongooseim3@localhost"
        ]
        ....
```

We need to ensure that unavailableNodes is a subset of discoveredNodes before returning it to the user.

### DIFF
--- a/test/cets_SUITE.erl
+++ b/test/cets_SUITE.erl
@@ -1091,12 +1091,11 @@ status_joined_nodes(Config) ->
         {{ok, [Node1, Node2]}, State}
     end,
     DiscoName = disco_name(Config),
-    Disco1 = start_disco(Node1, #{
+    DiscoOpts = #{
         name => DiscoName, backend_module => cets_discovery_fun, get_nodes_fn => F
-    }),
-    Disco2 = start_disco(Node2, #{
-        name => DiscoName, backend_module => cets_discovery_fun, get_nodes_fn => F
-    }),
+    },
+    Disco1 = start_disco(Node1, DiscoOpts),
+    Disco2 = start_disco(Node2, DiscoOpts),
     Tab = make_name(Config),
     {ok, _} = start(Node1, Tab),
     {ok, _} = start(Node2, Tab),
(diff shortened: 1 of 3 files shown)


test/cets_SUITE.erl
```diff
@@ -1054,19 +1056,46 @@ status_unavailable_nodes(Config) ->
...
+status_unavailable_nodes_is_subset_of_discovery_nodes(Config) ->
+    Node1 = node(),
+    Me = self(),
+    %% We use our process dictionary to set disco nodes dynamically
+    put(disco_nodes, [Node1, 'badnode@localhost']),
+    F = fun(State) ->
+        {dictionary, Dict} = erlang:process_info(Me, dictionary),
+        Nodes = proplists:get_value(disco_nodes, Dict),
+        {{ok, Nodes}, State}
+    end,
```
chrzaszcz: For me using `meck` would be more intuitive and easier to understand that we are mocking the discovery backend.
```suggestion
    meck:new(disco_backend, [non_strict]),
    meck:expect(disco_backend, get_disco_nodes, [Node1, 'badnode@localhost']),
```

test/cets_SUITE.erl
```diff
@@ -1054,19 +1056,46 @@ status_unavailable_nodes(Config) ->
...
+        name => DiscoName, backend_module => cets_discovery_fun, get_nodes_fn => F
+    }),
+    %% Disco needs at least one table to start calling get_nodes function
+    Tab = make_name(Config),
+    {ok, _} = start(Node1, Tab),
+    cets_discovery:add_table(Disco, Tab),
+    ok = cets_discovery:wait_for_ready(DiscoName, 5000),
+    ?assertMatch(#{unavailable_nodes := ['badnode@localhost']}, cets_status:status(DiscoName)),
+    %% Remove badnode from disco
+    put(disco_nodes, [Node1]),
```
chrzaszcz:
```suggestion
    meck:expect(disco_backend, get_disco_nodes, [Node1]),
```

chrzaszcz (approved):
Looks good - I only added minor style-related comments.

test/cets_SUITE.erl
arcusfelis: Sure, but because we run tests in parallel, module name would have to be generated from the testcase name.
We would have to use `make_name(Config)`, instead of disco_backend.

Because if we copy the testcase (like to test a similar situation), it would stop working.

arcusfelis:
added meck

chrzaszcz (approved):
:ok_hand:

PR #36, opened 2023-10-06
Add insert_new_or_lookup/2 function

arcusfelis:
`insert_new/2` is useful, but there could be a case, when we want to insert new or read existing record.
And there could be a race condition between these two events (i.e. this race condition is even in the default ets module).
Which means the code would have to implement this:

```erlang
-spec init_ram_key(ProposedKey) -> Result when
      ProposedKey :: mod_keystore:key(),
      Result :: {ok, ActualKey} | {error, init_ram_key_failed},
      ActualKey :: mod_keystore:key().
init_ram_key(ProposedKey) ->
    init_ram_key(ProposedKey, 1, 3).

%% Inserts new key or returns already inserted.
-spec init_ram_key(Key, TriedTimes, Retries) -> Result when
      Result :: {ok, Key} | {error, init_ram_key_failed},
      Key :: mod_keystore:key(),
      TriedTimes :: non_neg_integer(),
      Retries :: non_neg_integer().
init_ram_key(#key{id = Id, key = Key}, _, 0) ->
    ?LOG_ERROR(#{what => init_ram_key_failed, id => Id, key => Key}),
    {error, init_ram_key_failed};
init_ram_key(ProposedKey = #key{id = Id = {_, HostType}, key = PropKey}, N, Retries) ->
    Tab = table_name(HostType),
    case cets:insert_new(Tab, {Id, PropKey}) of
        true ->
            {ok, ProposedKey};
        false ->
            case ets:lookup(Tab, Id) of
                [{Id, Key}] ->
                    %% Return already inserted key
                    {ok, #key{id = Id, key = Key}};
                [] ->
                    ?LOG_WARNING(#{what => init_ram_key_retry,
                                   id => Id, key => PropKey, tried_times => N}),
                    init_ram_key(ProposedKey, N + 1, Retries - 1)
            end
   end.
 ```
 
 Retry mechanism. Would also require tests, would also need to decide when to stop to retry... Sounds too much work.
 
 This new function allows to rewrite the above code as:
 
 ```erlang
-spec init_ram_key(ProposedKey) -> Result when
      ProposedKey :: mod_keystore:key(),
      Result :: {ok, ActualKey} | {error, init_ram_key_failed},
      ActualKey :: mod_keystore:key().
init_ram_key(ProposedKey = #key{id = Id = {_, HostType}, key = PropKey}) ->
    Tab = table_name(HostType),
    case cets:insert_new_or_lookup(Tab, {Id, PropKey}) of
        {true, _} ->
            {ok, ProposedKey};
        {false, [{Id, Key}]} ->
            {ok, #key{id = Id, key = Key}}
   end.
 ```
 
 It is in case you want to write a testcase for the false condition. If you don't want to have any special logic in case of conflict (and no separation in the cover reports):
 
  ```erlang
-spec init_ram_key(ProposedKey) -> Result when
      ProposedKey :: mod_keystore:key(),
      Result :: {ok, ActualKey} | {error, init_ram_key_failed},
      ActualKey :: mod_keystore:key().
init_ram_key(ProposedKey = #key{id = Id = {_, HostType}, key = PropKey}) ->
    Tab = table_name(HostType),
    {_, [{Id, Key}]} = cets:insert_new_or_lookup(Tab, {Id, PropKey}),
    {ok, #key{id = Id, key = Key}}.
 ```

### DIFF
--- a/src/cets.erl
+++ b/src/cets.erl
@@ -23,6 +23,7 @@
     insert/2,
     insert_many/2,
     insert_new/2,
+    insert_new_or_read/2,
     insert_serial/2,
     delete/2,
     delete_many/2,
@@ -64,6 +65,7 @@
     insert/2,
     insert_many/2,
     insert_new/2,
+    insert_new_or_read/2,
     insert_serial/2,
     delete/2,
     delete_many/2,
@@ -109,6 +111,7 @@
     | {delete_many, [term()]}
     | {delete_objects, [term()]}
     | {insert_new, tuple()}
+    | {insert_new_or_read, tuple()}
     | {leader_op, op()}.
 -type remote_op() ::
     {remote_op, Op :: op(), From :: from(), AckPid :: ack_pid(), JoinRef :: join_ref()}.
@@ -118,6 +121,7 @@
 -type servers() :: ordsets:ordset(server_pid()).
 -type state() :: #{
     tab := table_name(),
+    keypos := pos_integer(),
     ack_pid := ack_pid(),
     join_ref := join_ref(),
     %% Updated by set_other_servers/2 function only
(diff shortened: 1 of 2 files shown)


chrzaszcz (approved):
:ok_hand:

PR #39, opened 2023-11-09
DNS related fixes

arcusfelis:
Check if DNS works fine before joining any nodes.


arcusfelis:
a different solution is introduced in https://github.com/esl/cets/pull/41 instead

PR #41, opened 2023-11-20
Rewrite ping

arcusfelis:
Changes:
- Introduce cets_ping with better node pinging strategy. It interrupts ping, if pang is received. It does not use net_adm:ping/1 to avoid going though net_kernel and potentially blocking it for a couple of seconds if the remote node is not available (couple of seconds adds up fast, if we have a lot of nodes). The current strategy is try to call EPMD first to check if the remote address is resolvable.
- Improve logging. Report nodeup/nodedown with downtime duration and time since the node start. It is especially useful when tweaking probes and upgrading k8s.
- Added node_down_history field to CETS. It looks like currently unused, but the purpose is to be able to analyse it manually and check after netsplits. It is added, because before we had conflict_nodes/conflict_tables persistently (i.e. after a netsplit, nodes see different lists). Still, I cannot reproduce conflict bug anymore, after tens of runs - so works as expected.

### DIFF
--- a/src/cets_ping.erl
+++ b/src/cets_ping.erl
@@ -15,8 +15,11 @@ ping(Node) when is_atom(Node) ->
             pong;
         false ->
             case dist_util:split_node(Node) of
-                {node, _Name, Host} ->
-                    case {inet:getaddr(Host, inet6), inet:getaddr(Host, inet)} of
+                {node, Name, Host} ->
+                    Epmd = net_kernel:epmd_module(),
+                    V4 = Epmd:address_please(Name, Host, inet),
+                    V6 = Epmd:address_please(Name, Host, inet6),
+                    case {V4, V6} of
                         {{error, _}, {error, _}} ->
                             pang;
                         _ ->
--- a/src/cets_ping.erl
+++ b/src/cets_ping.erl
@@ -40,12 +40,11 @@ net_family(_) ->
     inet.
 
 connect_ping(Node) ->
-    %% We could use net_adm:ping/1 but it does:
-    %% - disconnect node on pang - we don't want that
-    %%   (because it could disconnect already connected node because of race conditions)
-    %% - it calls net_kernel's gen_server of the remote server,
-    %%   but it could be busy doing something,
-    %%   which means slower response time.
+    %% We could use net_adm:ping/1 but it:
+    %% - disconnects node on pang - we don't want that
+    %%   (because it could disconnect an already connected node because of race conditions)
+    %% - calls net_kernel's gen_server of the remote server,
+    %%   but it could be busy doing something, which means slower response time.
     case net_kernel:connect_node(Node) of
         true ->
             pong;
@@ -69,7 +68,7 @@ ping_pairs_stop_on_pang([{Node1, Node2} | Pairs]) ->
         Other ->
             %% We do not need to ping the rest of nodes -
             %% one node returning pang is enough to cancel join.
-            %% We could exit earlier and safe some time
+            %% We could exit earlier and save some time
             %% (connect_node to the dead node could be time consuming)
             [{Node1, Node2, Other} | fail_pairs(Pairs, skipped)]
     end;
--- a/src/cets_ping.erl
+++ b/src/cets_ping.erl
@@ -19,17 +19,13 @@ ping(Node) when is_atom(Node) ->
         true ->
             pong;
         false ->
-            case dist_util:split_node(Node) of
-                {node, Name, Host} ->
-                    Epmd = net_kernel:epmd_module(),
-                    case Epmd:address_please(Name, Host, net_family()) of
-                        {error, _} ->
-                            pang;
-                        _ ->
-                            connect_ping(Node)
-                    end;
+            {node, Name, Host} = dist_util:split_node(Node),
+            Epmd = net_kernel:epmd_module(),
+            case Epmd:address_please(Name, Host, net_family()) of
+                {error, _} ->
+                    pang;
                 _ ->
-                    pang
+                    connect_ping(Node)
             end
     end.
 
--- a/src/cets_ping.erl
+++ b/src/cets_ping.erl
@@ -1,5 +1,10 @@
 -module(cets_ping).
 -export([ping/1, ping_pairs/1]).
+
+-ifdef(TEST).
+-export([net_family/1]).
+-endif.
+
 ping(Node) when is_atom(Node) ->
     %% It is important to understand, that initial setup for dist connections
     %% is done by the single net_kernel process.
--- a/src/cets_ping.erl
+++ b/src/cets_ping.erl
@@ -1,5 +1,5 @@
 -module(cets_ping).
--export([ping/1]).
+-export([ping/1, ping_pairs/1]).
 ping(Node) when is_atom(Node) ->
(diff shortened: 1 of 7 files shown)


src/cets_ping.erl
```diff
@@ -0,0 +1,86 @@
...
+net_family({ok, [[ProtoDist]]}) ->
+    %% Check that the string contains "6". i.e. inet6, inet6_tls.
+    case lists:member($6, ProtoDist) of
+        true ->
+            inet6;
+        false ->
+            inet
+    end;
+net_family(_) ->
+    inet.
```
chrzaszcz: This can be simplified, because the ones for IPv6 start with `inet6`:
```suggestion
net_family({ok, [["inet6" ++ _]]}) ->
    inet6;
net_family(_) ->
    inet.
```

src/cets_ping.erl
```diff
@@ -0,0 +1,86 @@
...
+
+connect_ping(Node) ->
+    %% We could use net_adm:ping/1 but it does:
+    %% - disconnect node on pang - we don't want that
+    %%   (because it could disconnect already connected node because of race conditions)
+    %% - it calls net_kernel's gen_server of the remote server,
+    %%   but it could be busy doing something,
+    %%   which means slower response time.
```
chrzaszcz:
```suggestion
    %% We could use net_adm:ping/1 but it:
    %% - disconnects node on pang - we don't want that
    %%   (because it could disconnect an already connected node because of race conditions)
    %% - calls net_kernel's gen_server of the remote server,
    %%   but it could be busy doing something, which means slower response time.
```

src/cets_ping.erl
```diff
@@ -0,0 +1,86 @@
...
+    F = fun() -> rpc:call(Node1, cets_ping, ping, [Node2], 10000) end,
+    Info = #{task => ping_node, node1 => Node1, node2 => Node2},
+    Res = cets_long:run_tracked(Info, F),
+    case Res of
+        pong ->
+            [{Node1, Node2, pong} | ping_pairs_stop_on_pang(Pairs)];
+        Other ->
+            %% We do not need to ping the rest of nodes -
+            %% one node returning pang is enough to cancel join.
+            %% We could exit earlier and safe some time
```
chrzaszcz:
```suggestion
            %% We could exit earlier and save some time
```

test/cets_SUITE.erl
```diff
@@ -1009,6 +1029,45 @@ atom_error_is_logged_in_tracked(_Config) ->
...
+        Me ! ready,
+        timer:sleep(infinity)
+    end,
+    Pid = spawn(fun() -> cets_long:run_tracked(#{log_ref => LogRef}, F) end),
+    receive_message(ready),
+    exit(Pid, shutdown),
+    wait_for_down(Pid),
+    [] = receive_all_logs_with_log_ref(?FUNCTION_NAME, LogRef).
+
+%% Counter example for shutdown_reason_is_not_logged_in_tracked
```
chrzaszcz: I'd rather change it to something like this:
```suggestion
%% Complementary to shutdown_reason_is_not_logged_in_tracked
```

test/cets_SUITE.erl
```diff
@@ -2208,10 +2264,340 @@ join_interrupted_when_ping_crashes(Config) ->
...
+    {ok, _Pid1} = start(Node1, Tab),
+    {ok, _Pid2} = start(Peer2, Tab),
+    F = fun(State) ->
+        {{ok, [Node1, Node2]}, State}
+    end,
+    DiscoName = disco_name(Config),
+    Disco = start_disco(Node1, #{
+        name => DiscoName, backend_module => cets_discovery_fun, get_nodes_fn => F
+    }),
+    cets_discovery:add_table(Disco, Tab),
```
chrzaszcz: Could you reuse this part, because it is the same as in the previous test case?

test/cets_SUITE.erl
```diff
@@ -2208,10 +2264,340 @@ join_interrupted_when_ping_crashes(Config) ->
...
+        }} = M ->
+            ?assert(is_integer(maps:get(alive_nodes, R)), M),
+            ?assert(is_integer(maps:get(time_since_startup_in_milliseconds, R)), M),
+            ?assert(is_integer(maps:get(connected_millisecond_duration, R)), M)
+    after 5000 ->
+        ct:fail(timeout)
+    end.
+
+disco_node_down_timestamp_is_remembered(Config) ->
+    logger_debug_h:start(#{id => ?FUNCTION_NAME}),
```
chrzaszcz: This line is probably unecessary

test/cets_SUITE.erl
```diff
@@ -2208,10 +2264,340 @@ join_interrupted_when_ping_crashes(Config) ->
...
+    F = fun(State) ->
+        {{ok, [Node1, Node2]}, State}
+    end,
+    DiscoName = disco_name(Config),
+    Disco = start_disco(Node1, #{
+        name => DiscoName, backend_module => cets_discovery_fun, get_nodes_fn => F
+    }),
+    cets_discovery:add_table(Disco, Tab),
+    ok = cets_discovery:wait_for_ready(Disco, 5000),
+    rpc(Peer2, erlang, disconnect_node, [Node1]),
```
chrzaszcz: This part also repeats - I had to spend some time on comparing the two parts, and with a helper function it would be obvious.

test/cets_SUITE.erl
```diff
@@ -2208,10 +2264,340 @@ join_interrupted_when_ping_crashes(Config) ->
...
```
chrzaszcz: Why disconnect if we are checking `nodeup`, not `nodedown`? Is this line unnecessary? It looks like it was copied from the testcase below.

test/cets_SUITE.erl
```diff
@@ -2208,10 +2264,340 @@ join_interrupted_when_ping_crashes(Config) ->
...
+    DiscoName = disco_name(Config),
+    Disco = start_disco(Node1, #{
+        name => DiscoName, backend_module => cets_discovery_fun, get_nodes_fn => F
+    }),
+    %% We have to start discovery server on both nodes for that feature to work
+    Disco2 = start_disco(Node2, #{
+        name => DiscoName, backend_module => cets_discovery_fun, get_nodes_fn => F
+    }),
+    cets_discovery:add_table(Disco, Tab),
+    ok = cets_discovery:wait_for_ready(Disco, 5000),
```
chrzaszcz: Unless I missed something (it's difficult to compare manually), this chunk repeats for 3 test cases. Could you extract it to a helper?

chrzaszcz (commented):
Looks good in general. The test suite is becoming lengthy, and I was getting lost in the test cases...

test/cets_SUITE.erl
arcusfelis: We check reconnect here.
After disconnect_node on this line, disco should reconnect automatically.
We selectively receive nodeup with downtime_millisecond_duration field.
downtime_millisecond_duration field would only get added after the second nodeup (first one we never extract from the message box).

If we remove this disconnect_node line, the test would fail with the timeout in the receive statement.

I add more comments here into the code though, describing what we test.

chrzaszcz (approved):
👌

PR #42, opened 2023-11-24
Add cets_discovery:wait_for_get_nodes/2

arcusfelis:
This function blocks until get_nodes returns

### DIFF
--- a/src/cets_discovery.erl
+++ b/src/cets_discovery.erl
@@ -506,7 +506,7 @@ handle_nodedown(Node, State) ->
         set_defined(connected_millisecond_duration, NodeUpTime, #{
             what => nodedown,
             remote_node => Node,
-            alive_nodes => length(nodes()) + 1,
+            connected_nodes => length(nodes()) + 1,
             time_since_startup_in_milliseconds => time_since_startup_in_milliseconds(State)
         })
     ),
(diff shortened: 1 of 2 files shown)


src/cets_discovery.erl
```diff
@@ -146,14 +149,24 @@ info(Server) ->
...
-%% It also waits till the data is loaded from the remote nodes
+%% This calls blocks until the initial discovery is done.
+%% It also waits till the data is loaded from the remote nodes.
 -spec wait_for_ready(server(), timeout()) -> ok.
 wait_for_ready(Server, Timeout) ->
     F = fun() -> gen_server:call(Server, wait_for_ready, Timeout) end,
     Info = #{task => cets_wait_for_ready},
     cets_long:run_tracked(Info, F).
 
+%% Waits for the currect get_nodes call to return.
```
chrzaszcz:
```suggestion
%% Waits for the current get_nodes call to return.
```

chrzaszcz (approved):
Looks good. There is a minor typo.

chrzaszcz (approved):
The added function waits for the `get_nodes` function to finish. It is difficult to say if it fixes the race conditions, because it is not described what it is supposed to fix. I can approve, because it works as described. We will see how it fits into MongooseIM.

PR #43, opened 2023-11-28
Enable cets_discovery get_nodes queries even with no tables

arcusfelis:
Enable cets_discovery get_nodes queries even with no tables.

Discovery is important to join nodes, so enable it even with no added tables to track.
This should simplify logic and testing a bit.


chrzaszcz (approved):
👍

PR #45, opened 2023-12-04
Fix unavailableNodes list having joined nodes

arcusfelis:
Ping result with pang could arrive after nodeup

Fix for wrong list in unavailableNodes :

```
Result 0 {
  "data" : {
    "cets" : {
      "systemInfo" : {
        "unavailableNodes" : [ // these should not be here
          "mongooseim@mongooseim-1.mongooseim.default.svc.cluster.local",
          "mongooseim@mongooseim-2.mongooseim.default.svc.cluster.local"
        ],
        "remoteUnknownTables" : [
          
        ],
        "remoteNodesWithoutDisco" : [
          
        ],
        "remoteNodesWithUnknownTables" : [
          
        ],
        "remoteNodesWithMissingTables" : [
          
        ],
        "remoteMissingTables" : [
          
        ],
        "joinedNodes" : [
          "mongooseim@mongooseim-0.mongooseim.default.svc.cluster.local",
          "mongooseim@mongooseim-1.mongooseim.default.svc.cluster.local",
          "mongooseim@mongooseim-2.mongooseim.default.svc.cluster.local"
        ],
        "discoveryWorks" : true,
        "discoveredNodes" : [
          "mongooseim@mongooseim-0.mongooseim.default.svc.cluster.local",
          "mongooseim@mongooseim-1.mongooseim.default.svc.cluster.local",
          "mongooseim@mongooseim-2.mongooseim.default.svc.cluster.local"
        ],
        "conflictTables" : [
          
        ],
        "conflictNodes" : [
          
        ],
        "availableNodes" : [
          "mongooseim@mongooseim-0.mongooseim.default.svc.cluster.local",
          "mongooseim@mongooseim-1.mongooseim.default.svc.cluster.local",
          "mongooseim@mongooseim-2.mongooseim.default.svc.cluster.local"
        ]
      }
    }
  }
}
```

Tested with:
https://github.com/esl/MongooseIM/pull/4185

Tested with Helm tests.

### DIFF
--- a/test/cets_SUITE.erl
+++ b/test/cets_SUITE.erl
@@ -191,6 +191,7 @@ seq_cases() ->
         disco_node_down_timestamp_is_remembered,
         disco_nodeup_timestamp_is_updated_after_node_reconnects,
         disco_node_start_timestamp_is_updated_after_node_restarts,
+        disco_pang_result_arrives_after_nodeup,
         disco_nodeup_triggers_check_and_get_nodes,
         ping_pairs_returns_pongs,
         ping_pairs_returns_earlier,
@@ -205,7 +206,8 @@ cets_seq_no_log_cases() ->
         disco_node_up_timestamp_is_remembered,
         disco_node_down_timestamp_is_remembered,
         disco_nodeup_timestamp_is_updated_after_node_reconnects,
-        disco_node_start_timestamp_is_updated_after_node_restarts
+        disco_node_start_timestamp_is_updated_after_node_restarts,
+        disco_pang_result_arrives_after_nodeup
     ].
 
 init_per_suite(Config) ->
@@ -2471,6 +2473,19 @@ disco_node_start_timestamp_is_updated_after_node_restarts(Config) ->
     simulate_disco_restart(Setup),
     wait_for_disco_timestamp_to_be_updated(Disco, node_start_timestamps, Node2, OldTimestamp).
 
+disco_pang_result_arrives_after_nodeup(Config) ->
+    meck:new(cets_ping, [passthrough]),
+    meck:expect(cets_ping, send_ping_result, fun(Pid, Node, _PingResult) ->
+        meck:passthrough([Pid, Node, pang])
+    end),
+    try
+        Setup = setup_two_nodes_and_discovery(Config, [wait]),
+        #{disco_name := DiscoName} = Setup,
+        ?assertMatch([], maps:get(unavailable_nodes, cets_status:status(DiscoName)))
+    after
+        meck:unload()
+    end.
+
 disco_nodeup_triggers_check_and_get_nodes(Config) ->
     Setup = setup_two_nodes_and_discovery(Config, [wait, notify_get_nodes]),
     #{disco := Disco, node2 := Node2} = Setup,
@@ -2743,12 +2758,17 @@ given_n_servers(Config, N, Opts) ->
 setup_two_nodes_and_discovery(Config) ->
     setup_two_nodes_and_discovery(Config, []).
 
+%% Flags:
+%% - disco2 - start discovery on Node2
+%% - wait - call wait_for_ready/2
 setup_two_nodes_and_discovery(Config, Flags) ->
+    ok = net_kernel:monitor_nodes(true),
     Me = self(),
     Node1 = node(),
     #{ct2 := Peer2} = proplists:get_value(peers, Config),
     #{ct2 := Node2} = proplists:get_value(nodes, Config),
     disconnect_node(Peer2, Node1),
+    receive_message({nodedown, Node2}),
     Tab = make_name(Config),
     {ok, _Pid1} = start(Node1, Tab),
     {ok, _Pid2} = start(Peer2, Tab),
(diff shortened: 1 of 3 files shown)


test/cets_SUITE.erl
```diff
@@ -2471,6 +2473,19 @@ disco_node_start_timestamp_is_updated_after_node_restarts(Config) ->
     simulate_disco_restart(Setup),
     wait_for_disco_timestamp_to_be_updated(Disco, node_start_timestamps, Node2, OldTimestamp).
 
+disco_pang_result_arrives_after_nodeup(Config) ->
```
chrzaszcz: Please make the test case name consistent with what it does. It says that `pang` arrives after `nodeup`, but instead we just always return `pang`. Also, the test could be misleading, because it looks like we expect everything to be OK despite ping result always being `pang`.

chrzaszcz (commented):
Looks good in general. I think that node state could be simplified later because of too many lists in the status - one list with node status for each node instead.

test/cets_SUITE.erl
```diff
@@ -2473,15 +2473,38 @@ disco_node_start_timestamp_is_updated_after_node_restarts(Config) ->
...
+    %% So, there could be the situation when the result of ping would be processed
+    %% after the node actually got connected.
     meck:new(cets_ping, [passthrough]),
+    Me = self(),
     meck:expect(cets_ping, send_ping_result, fun(Pid, Node, _PingResult) ->
+        %% Wait until Node is up
+        Cond = fun() -> lists:member(Node, nodes()) end,
+        cets_test_wait:wait_until(Cond, true),
+        Me ! send_ping_result_called,
+        %% Similate pang returned to cets_discovery.
```
chrzaszcz: Simulate? Simulated? I can't understand this sentence.

test/cets_SUITE.erl
```diff
@@ -2473,15 +2473,38 @@ disco_node_start_timestamp_is_updated_after_node_restarts(Config) ->
...
+        %% cets_join does not use send_ping_result function
+        %% and would receive pong and join correctly.
         meck:passthrough([Pid, Node, pang])
     end),
     try
-        Setup = setup_two_nodes_and_discovery(Config, [wait]),
+        %% setup_two_nodes_and_discovery would call disconnect_node node
```
chrzaszcz: "disconnect_node node"? Can't understand it either. Is it a typo or did you mean a particular node?

test/cets_SUITE.erl
```diff
@@ -2473,15 +2473,38 @@ disco_node_start_timestamp_is_updated_after_node_restarts(Config) ->
...
         #{disco_name := DiscoName} = Setup,
-        ?assertMatch([], maps:get(unavailable_nodes, cets_status:status(DiscoName)))
+        Status = cets_status:status(DiscoName),
+        %% Check that pang is ignored and unavailable_nodes list is empty.
+        ?assertMatch([], maps:get(unavailable_nodes, Status)),
+        %% Ensure that Node2 is in a list of joined nodes
+        ?assertMatch([Node1, Node2], maps:get(joined_nodes, Status)),
+        %% Ensure that send_ping_result was called
+        %% (it is not called if node is in the nodes() list)
+        receive_message(send_ping_result_called)
```
chrzaszcz: I doesn't make sense to check it after the testcase, because it might have arrived after we checked unavailable nodes. I would move it before all the `?assertMatch`

test/cets_SUITE.erl
```diff
@@ -2473,15 +2473,38 @@ disco_node_start_timestamp_is_updated_after_node_restarts(Config) ->
...
+        %% Similate pang returned to cets_discovery.
+        %% cets_join does not use send_ping_result function
+        %% and would receive pong and join correctly.
         meck:passthrough([Pid, Node, pang])
     end),
     try
-        Setup = setup_two_nodes_and_discovery(Config, [wait]),
+        %% setup_two_nodes_and_discovery would call disconnect_node node
+        %% before connecting it back.
+        Setup = setup_two_nodes_and_discovery(Config, [wait, disco2]),
```
chrzaszcz, line 2498: I still don't understand how we force `pang` to arrive after the node went up. If it's a race condition, how we are enforcing it here?

test/cets_SUITE.erl
```diff
@@ -2471,6 +2473,42 @@ disco_node_start_timestamp_is_updated_after_node_restarts(Config) ->
...
+        #{disco_name := DiscoName} = Setup,
+        Status = cets_status:status(DiscoName),
+        %% Check that pang is ignored and unavailable_nodes list is empty.
+        ?assertMatch([], maps:get(unavailable_nodes, Status)),
+        %% Ensure that Node2 is in a list of joined nodes
+        ?assertMatch([Node1, Node2], maps:get(joined_nodes, Status)),
+        %% Ensure that send_ping_result was called
```
chrzaszcz: This comment is unnecessary because it just says `receive_message(send_ping_result_called)` only in different words.

test/cets_SUITE.erl
```diff
@@ -2471,6 +2473,42 @@ disco_node_start_timestamp_is_updated_after_node_restarts(Config) ->
...
+        %% (it is not called if node is in the nodes() list)
```
chrzaszcz: ...but this one I don't get at all. Which node? And previous line says "ensure it's called" while this one says "it is not called"... very confusing.

chrzaszcz (commented):
I added a few more comments... I really think this case got overly verbose and it's hard to understand some of the comments. I would leave only the ones that are most important, i.e. the ones that explain when the node is up, and that we send `pang` afterwards.

test/cets_SUITE.erl
arcusfelis: it makes some sense, because here is how the function looks like:
```
ping_not_connected_nodes(Nodes) ->
    Self = self(),
    NotConNodes = Nodes -- [node() | nodes()],
    [
        spawn_link(fun() -> cets_ping:send_ping_result(Self, Node, cets_ping:ping(Node)) end)
     || Node <- lists:sort(NotConNodes)
    ],
    ok.
 ```
 
 So there is a chance that node got connected by the reason not related to the current test - in this case `cets_ping:send_ping_result/2` would never be called :)

chrzaszcz (approved):
Looks better and more readable now, fine for me 👍

PR #46, opened 2023-12-23
Pause on all nodes during join

arcusfelis:
To avoid race conditions we have to pause on all nodes.

The issue:
- if join coordinator node looses connection with one of the nodes during join, that node gets unpaused.
- when unpaused, the node would start sending remote ops and check_server requests.
- there is a chance that other nodes in cluster would receive the messages before send_dump.

I've tried to show the errors in tests.

Solution:
- we could delay the messages by unpausing only when all nodes receive DOWN from the cets_join process.

Adresses MIM-2137

### DIFF
--- a/src/cets.erl
+++ b/src/cets.erl
@@ -119,6 +119,8 @@
 -type table_name() :: atom().
 -type pause_monitor() :: reference().
 -type servers() :: ordsets:ordset(server_pid()).
+-type delayed_check_server() :: {FromPid :: pid(), JoinRef :: join_ref()}.
+-type node_down_history() :: #{node => node(), pid => pid(), reason => term()}.
 -type state() :: #{
     tab := table_name(),
     keypos := pos_integer(),
@@ -131,7 +133,8 @@
     opts := start_opts(),
     backlog := [backlog_entry()],
     pause_monitors := [pause_monitor()],
-    node_down_history := [node()]
+    delayed_check_server := [delayed_check_server()],
+    node_down_history := [node_down_history()]
 }.
 
 -type long_msg() ::
@@ -157,7 +160,8 @@
     ack_pid := ack_pid(),
     join_ref := join_ref(),
     opts := start_opts(),
-    node_down_history := [node()]
+    node_down_history := [node_down_history()],
+    paused := boolean()
 }.
 
 -type handle_down_fun() :: fun((#{remote_pid := server_pid(), table := table_name()}) -> ok).
@@ -231,7 +235,7 @@ table_name(Tab) when is_atom(Tab) ->
 table_name(Server) ->
     cets_call:long_call(Server, table_name).
 
--spec send_dump(server_ref(), servers(), join_ref(), [tuple()]) -> ok.
+-spec send_dump(server_ref(), servers(), join_ref(), [tuple()]) -> ok | {error, ignored}.
 send_dump(Server, NewPids, JoinRef, OurDump) ->
     Info = #{msg => send_dump, join_ref => JoinRef, count => length(OurDump)},
     cets_call:long_call(Server, {send_dump, NewPids, JoinRef, OurDump}, Info).
@@ -421,6 +425,7 @@ init({Tab, Opts}) ->
         opts => Opts,
         backlog => [],
         pause_monitors => [],
+        delayed_check_server => [],
         node_down_history => []
     }}.
 
@@ -487,8 +492,7 @@ handle_info({remote_op, Op, From, AckPid, JoinRef}, State) ->
 handle_info({'DOWN', Mon, process, Pid, Reason}, State) ->
     {noreply, handle_down(Mon, Pid, Reason, State)};
 handle_info({check_server, FromPid, JoinRef}, State) ->
-    handle_check_server(FromPid, JoinRef, State),
-    {noreply, State};
+    {noreply, handle_check_server(FromPid, JoinRef, State)};
 handle_info(Msg, State) ->
     ?LOG_ERROR(#{what => unexpected_info, msg => Msg}),
     {noreply, State}.
@@ -502,6 +506,14 @@ code_change(_OldVsn, State, _Extra) ->
 %% Internal logic
 
 -spec handle_send_dump(servers(), join_ref(), [tuple()], state()) -> {reply, ok, state()}.
+handle_send_dump(_NewPids, JoinRef, _Dump, State = #{pause_monitors := []}) ->
+    ?LOG_ERROR(#{
+        what => send_dump_received_when_unpaused,
+        text => <<"Received send_dump message while in the unpaused state. Ignore it">>,
+        join_ref => JoinRef,
+        state => State
+    }),
+    {reply, {error, ignored}, State};
 handle_send_dump(NewPids, JoinRef, Dump, State = #{tab := Tab, other_servers := Servers}) ->
     ets:insert(Tab, Dump),
     Servers2 = add_servers(NewPids, Servers),
@@ -519,17 +531,17 @@ handle_down(Mon, Pid, Reason, State = #{pause_monitors := Mons}) ->
             }),
             handle_unpause2(Mon, Mons, State);
         false ->
-            handle_down2(Pid, State)
+            handle_down2(Pid, Reason, State)
     end.
 
--spec handle_down2(pid(), state()) -> state().
-handle_down2(RemotePid, State = #{other_servers := Servers, ack_pid := AckPid}) ->
+-spec handle_down2(pid(), term(), state()) -> state().
+handle_down2(RemotePid, Reason, State = #{other_servers := Servers, ack_pid := AckPid}) ->
     case lists:member(RemotePid, Servers) of
         true ->
             cets_ack:send_remote_down(AckPid, RemotePid),
             call_user_handle_down(RemotePid, State),
             Servers2 = lists:delete(RemotePid, Servers),
-            update_node_down_history(RemotePid, set_other_servers(Servers2, State));
+            update_node_down_history(RemotePid, Reason, set_other_servers(Servers2, State));
         false ->
             %% This should not happen
             ?LOG_ERROR(#{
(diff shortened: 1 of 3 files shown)


src/cets.erl
```diff
@@ -119,6 +119,7 @@
 -type table_name() :: atom().
 -type pause_monitor() :: reference().
 -type servers() :: ordsets:ordset(server_pid()).
+-type node_down_history() :: #{node => node(), pid => pid(), reason => term()}.
```
chrzaszcz: I'd name this one `node_down_event` or `node_down_item`, because it is not the whole history (which is a list).

test/cets_SUITE.erl
```diff
@@ -82,6 +85,7 @@ cases() ->
         remote_ops_are_ignored_if_join_ref_does_not_match,
         join_retried_if_lock_is_busy,
         send_dump_contains_already_added_servers,
+        servers_remove_each_other_each_other_if_join_refs_do_not_match_after_unpause,
```
chrzaszcz:
```suggestion
        servers_remove_each_other_if_join_refs_do_not_match_after_unpause,
```

test/cets_SUITE.erl
```diff
@@ -1135,10 +1150,184 @@ send_dump_contains_already_added_servers(Config) ->
...
+                _ -> false
+            end
+        end,
+        true
+    ),
+    {monitors, [{process, MonitorProcess}]} = erlang:process_info(TestPid, monitors),
+    exit(MonitorProcess, killed),
+    wait_for_down(TestPid).
+
+pause_on_remote_crashes(Config) ->
```
chrzaszcz:
```suggestion
pause_on_remote_node_crashes(Config) ->
```

test/cets_SUITE.erl
```diff
@@ -1135,10 +1150,184 @@ send_dump_contains_already_added_servers(Config) ->
...
+                {assert_all_ok, [
+                    {Node2, {error, {exception, mock_pause_on_remote_node_failing, _}}}
+                ]},
+                #{}}} =
+            cets_join:join(lock_name(Config), #{}, Pid1, Pid2, #{})
+    after
+        [cets_join] = rpc(Node2, meck, unload, [])
+    end.
+
+%% Happens when one node receives send_dump and looses connection with the node
```
chrzaszcz:
```suggestion
%% Happens when one node receives send_dump and loses connection with the node
```

test/cets_SUITE.erl
```diff
@@ -1135,10 +1150,184 @@ send_dump_contains_already_added_servers(Config) ->
...
+    #{ct6 := Peer6, ct7 := Peer7, ct5 := Peer5} = proplists:get_value(peers, Config),
+    Tab = make_name(Config),
+    Lock = lock_name(Config),
+    {ok, Pid6} = start(Peer6, Tab),
+    {ok, Pid7} = start(Peer7, Tab),
+    CheckPointF = fun
+        ({before_send_dump, Pid}) when Pid == Pid7 ->
+            %% Node6 already got its dump
+            disconnect_node(Peer6, node()),
+            %% Wait for Pid6 to loose pause monitor from the join process
```
chrzaszcz:
```suggestion
            %% Wait for Pid6 to lose pause monitor from the join process
```

test/cets_SUITE.erl
```diff
@@ -1135,10 +1150,184 @@ send_dump_contains_already_added_servers(Config) ->
...
+    #{ct6 := Peer6, ct7 := Peer7, ct5 := Peer5} = proplists:get_value(peers, Config),
+    Tab = make_name(Config),
+    Lock = lock_name(Config),
+    {ok, Pid6} = start(Peer6, Tab),
+    {ok, Pid7} = start(Peer7, Tab),
+    insert(Peer6, Tab, {a, 1}),
+    CheckPointF = fun
+        ({before_send_dump, Pid}) when Pid == Pid7 ->
+            %% Node6 already got its dump.
+            %% Use disconnect_node to loose connection between the coordinator and Pid6.
```
chrzaszcz:
```suggestion
            %% Use disconnect_node to lose connection between the coordinator and Pid6.
```

src/cets_join.erl
```diff
@@ -135,10 +140,51 @@ join2(Info, LocalPid, RemotePid, JoinOpts) ->
         lists:foreach(fun({Pid, Ref}) -> catch cets:unpause(Pid, Ref) end, Paused)
     end.
 
-send_dump(Pid, Pids, JoinRef, Dump, JoinOpts) ->
+-spec pause_servers(AllPids :: [pid(), ...]) -> Paused :: [{pid(), cets:pause_monitor()}].
+pause_servers(AllPids) ->
+    %% We should create a pause helper process on each node in the cluster.
+    %% It is to ensure that node that loosing a connection with cets_join coordinator
```
chrzaszcz:
```suggestion
    %% It is to ensure that node that losing a connection with cets_join coordinator
```

src/cets_join.erl
```diff
@@ -135,10 +140,51 @@ join2(Info, LocalPid, RemotePid, JoinOpts) ->
...
+
+-spec pause_on_remote_node(pid(), [pid()]) -> ok.
+pause_on_remote_node(JoinerPid, AllPids) ->
+    Self = self(),
+    {Pid, Mon} = spawn_monitor(fun() ->
+        JoinerMon = erlang:monitor(process, JoinerPid),
+        MyNode = node(),
+        %% Ignore pids on the current nodes
+        %% (because we only interested in internode connections here).
+        %% Catching because we can ignore loosing some connections here.
```
chrzaszcz:
```suggestion
        %% Catching because we can ignore losing some connections here.
```

src/cets_join.erl
```diff
@@ -135,10 +140,51 @@ join2(Info, LocalPid, RemotePid, JoinOpts) ->
...
+    assert_all_ok(OtherNodes, Results),
+    Paused.
+
+-spec pause_on_remote_node(pid(), [pid()]) -> ok.
+pause_on_remote_node(JoinerPid, AllPids) ->
+    Self = self(),
+    {Pid, Mon} = spawn_monitor(fun() ->
+        JoinerMon = erlang:monitor(process, JoinerPid),
+        MyNode = node(),
+        %% Ignore pids on the current nodes
```
chrzaszcz: Only one node is ignored, not some "current nodes"

src/cets_join.erl
```diff
@@ -135,10 +140,51 @@ join2(Info, LocalPid, RemotePid, JoinOpts) ->
         lists:foreach(fun({Pid, Ref}) -> catch cets:unpause(Pid, Ref) end, Paused)
     end.
 
-send_dump(Pid, Pids, JoinRef, Dump, JoinOpts) ->
+-spec pause_servers(AllPids :: [pid(), ...]) -> Paused :: [{pid(), cets:pause_monitor()}].
+pause_servers(AllPids) ->
+    %% We should create a pause helper process on each node in the cluster.
+    %% It is to ensure that node that loosing a connection with cets_join coordinator
+    %% would not unpause one of the processes too soon
```
chrzaszcz, line 147: Why would losing a connection unpause anything? Also: what is a join coordinator exactly? I don't see this term appearing anywhere else in the code.

src/cets_join.erl
```diff
@@ -135,10 +140,51 @@ join2(Info, LocalPid, RemotePid, JoinOpts) ->
...
+    %% (because it could start sending remote ops to nodes which are still in the current joining procedure).
+    Paused = [{Pid, cets:pause(Pid)} || Pid <- AllPids],
+    OtherNodes = lists:delete(node(), lists:usort([node(Pid) || Pid <- AllPids])),
+    Results = erpc:multicall(
+        OtherNodes, ?MODULE, pause_on_remote_node, [self(), AllPids], timer:seconds(30)
```
chrzaszcz, line 152: I don't get it - why do we make all nodes call pause on all nodes? It results in a quadratic number of pauses - each node would get paused N-1 times when N is the total number of nodes. It looks that you want to establish monitors between all node pairs, but it's just my guess.

chrzaszcz (commented):
I reviewed what I could understand. The test suite becomes huge and I have no idea how you don't get lost in the test cases - I did. I still don't get the cause of the original issue though, so I really can't evaluate the fix.

src/cets_join.erl
arcusfelis: because pause is a monitor.
It is done the way that if the pause caller dies, the pause is unpaused (because DOWN is received).

If there is a netsplit - a DOWN is sent too.

Join coordinator is a process/node that calls cets_join.

src/cets_join.erl
arcusfelis: kinda, would lead to 10k pauses if there are 100 nodes.
Pause on all nodes is a stronger way to do the current pause logic.
It is just the simplest solution to ensure that we start processing messages only when all connected to us nodes removed their pause claim.


The failure case is a bit tricky here - that is why test is hard to understand and that is why there is a bug to begin with. Probably, because my initial assumption was that just calling cets:pause/1 for all servers is enough to stop servers from interfering with the join procedure. But, apparently, if there is a netsplit, the disconnected node would get unpaused while still keeping a connection to other nodes, which are still not fully finished the join procedure.


Also, there is logic that: "After finishing join (i.e. after a server has 0 active pauses) we compare join_ref and: if join_ref matches, we start sending writes, if join_ref does not match - we remove that server from a list of servers to replicate data from". And for this logic to be functional it is critical that pauses remain working until the `cets_join:join/2` funtion returns.



Alternative solution would be to also put incoming remote_ops and `check_server` messages into a queue, if the server is paused. But it sounds like more mental overload and new types of race conditions possible.

arcusfelis: Basically, main issue is that if node disconnects for some reason, it does not disappears from all nodes atomically or even fast. So, there could be a short period of time when node looses connection with the node that executing cets_join:join, but still keeps a connection to other nodes to the cluster.

chrzaszcz:
After a discussion with @arcusfelis I understand the current solution, and it seems good enough for now. We just need to address the remaining comments.

chrzaszcz (approved):
Looks good 👍

PR #47, opened 2024-01-03
Add docs

arcusfelis:
Docs generated with:

```
rebar3 docs
```

in the `doc` directory.

### DIFF
--- a/rebar.config
+++ b/rebar.config
@@ -23,12 +23,6 @@
     {cover_tests, [
         {ct, "--sname=ct1 --cover"}
     ]},
-    {docs, [
-        {do, "ex_doc"}
-    ]},
-    {publish_to_hex, [
-        {do, "default as publish_to_hex hex publish --revert 0.0.1"}
-    ]},
     {all, [
         format,
         run_tests,
@@ -52,7 +46,7 @@
     ]}
 ]}.
 
-{plugins, [rebar3_ex_doc, hex]}.
+{plugins, [rebar3_ex_doc, rebar3_hex]}.
 
 {ex_doc, [
     {source_url, <<"https://github.com/esl/cets">>},
--- a/rebar.config
+++ b/rebar.config
@@ -24,7 +24,10 @@
         {ct, "--sname=ct1 --cover"}
     ]},
     {docs, [
-        {do, "default as docs ex_doc"}
+        {do, "ex_doc"}
+    ]},
+    {publish_to_hex, [
+        {do, "default as publish_to_hex hex publish --revert 0.0.1"}
     ]},
     {all, [
         format,
@@ -35,9 +38,6 @@
 ]}.
 
 {profiles, [
-    {docs, [
-        {plugins, [rebar3_ex_doc]}
-    ]},
     {format, [
         {plugins, [erlfmt]}
     ]},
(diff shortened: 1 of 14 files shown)


src/cets_long.erl
```diff
@@ -1,24 +1,30 @@
...
-%% Extra logging information
 -type log_info() :: map().
+%% Extra logging information.
+
 -type task_result() :: term().
+%% The generic result of execution.
+
 -type task_fun() :: fun(() -> task_result()).
--export_type([log_info/0]).
+%% User provided function to executre.
```
NelsonVides:
```suggestion
%% User provided function to execute.
```

NelsonVides (changes requested):
One thing is missing: https://github.com/esl/cets/blob/main/src/cets.app.src

Compare for example to https://github.com/esl/amoc/blob/master/src/amoc.app.src

We need to provide there `{vsn, git}`, license and links.

rebar.config
```diff
@@ -23,6 +23,12 @@
     {cover_tests, [
         {ct, "--sname=ct1 --cover"}
     ]},
+    {docs, [
+        {do, "ex_doc"}
+    ]},
+    {publish_to_hex, [
+        {do, "default as publish_to_hex hex publish --revert 0.0.1"}
+    ]},
```
NelsonVides: Please remove this

rebar.config
```diff
@@ -23,6 +23,12 @@
     {cover_tests, [
         {ct, "--sname=ct1 --cover"}
     ]},
+    {docs, [
+        {do, "ex_doc"}
+    ]},
```
NelsonVides: Why do we need this alias to begin with? Why can't we just do `rebar3 ex_doc` like everyone else do?

rebar.config
```diff
@@ -45,3 +51,13 @@
         ]}
     ]}
 ]}.
+
+{plugins, [rebar3_ex_doc, hex]}.
```
NelsonVides: It's called `rebar3_hex`, not `hex`: https://hex.pm/packages/rebar3_hex

src/cets.app.src
```diff
@@ -1,9 +1,12 @@
...
+    {vsn, git},
     {modules, []},
     {registered, []},
     {applications, [kernel, stdlib]},
-    {env, []}
-    % {mod, {cets_app, []}
+    {env, []},
+    {maintainers, ["ESL"]},
+    {pkg_name, "cets"},
+    {licenses, ["Apache License 2.0"]},
```
NelsonVides: License should be in the accepted formatting, see: https://spdx.org/licenses/
```suggestion
    {licenses, ["Apache-2.0"]},
```

src/cets.app.src
```diff
@@ -1,9 +1,12 @@
...
-    {vsn, "0.1"},
+    {vsn, git},
     {modules, []},
     {registered, []},
     {applications, [kernel, stdlib]},
-    {env, []}
-    % {mod, {cets_app, []}
+    {env, []},
+    {maintainers, ["ESL"]},
+    {pkg_name, "cets"},
```
NelsonVides: This is not needed, the default is already the name of the application.

NelsonVides (changes requested):

src/cets.app.src
arcusfelis: fullname works too, **actually**
(because Apache 2.0 would NOT work when publishing, but full name will).
Where is info that there should be IDs?

NelsonVides (approved):
LGTM 👌🏽

PR #48, opened 2024-01-05
Ignore pause_owner_crashed log message if the pause owner exits normally

arcusfelis:
MIM-2145

`pause_owner_crashed` log message informs the admin, that something is wrong with the code that called `cets:pause/1` function.

Pause is unpaused when:
- the pause caller exits.
- `cets:unpause/2` is called.

But recent changes for `cets_join` now use exit with reason=normal.

Changes:
- Ignore pause_owner_crashed.
- New helper function for tests `receive_all_logs_from_pid`.
- Move logger helper functions into a separate test helper module.

This just reduce number of logged messages, no changes to the join logic.


chrzaszcz (approved):
Looks good 👍

PR #50, opened 2024-01-31
Do not run check in cets_discovery on nodedown

arcusfelis:
Patches for disco.

- Do not call check on nodedown. This prevents immediate reconnect (which could confuse code inside global, trying to execute logic for prevent_overlapping_partitions).

- If nodedown is received, wait 30 second before the next check. This allows the restarting node to initiate the connection on its own, without the whole cluster trying to connect to it.

- Improve logging for when wait_for_ready fails.

- Provide extra args into handle_down_fun: is_leader and remote_node

### DIFF
--- a/src/cets_discovery.erl
+++ b/src/cets_discovery.erl
@@ -107,7 +107,7 @@
     phase := initial | regular,
     results := [join_result()],
     nodes := ordsets:ordset(node()),
-    %% The nodes that returned pang, sorted
+    %% The nodes that returned pang or nodedown, sorted
     unavailable_nodes := ordsets:ordset(node()),
     tables := [atom()],
     backend_module := module(),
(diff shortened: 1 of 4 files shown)


src/cets_discovery.erl
```diff
@@ -633,8 +657,15 @@ set_defined(Key, Value, Map) ->
 time_since_startup_in_milliseconds(#{start_time := StartTime}) ->
     time_since(StartTime).
 
+-spec time_since(integer()) -> integer().
 time_since(StartTime) ->
```
chrzaszcz:
```suggestion
time_since(StartTime) when is_integer(StartTime) ->
```

src/cets_discovery.erl
```diff
@@ -633,8 +657,15 @@ set_defined(Key, Value, Map) ->
...
-    erlang:system_time(millisecond) - StartTime.
+    %% Dialyzer thinks integer() - integer() could be float.
+    %% Do round to avoid the warning.
+    round(get_time() - StartTime).
```
chrzaszcz:
```suggestion
    get_time() - StartTime.
```

chrzaszcz (commented):
Looks good, but CI failed. I added a minor comment.

chrzaszcz (approved):
Looks good in general, but `cets_SUITE` is becoming huge, and for me it's really hard to follow the changes there. IMO we need to extract some helpers to separate modules - especially that this suite is full of generic helpers. It should be a separate story - please don't extend this PR.

PR #51, opened 2024-03-06
Dist blocker

arcusfelis:
In this PR we introduce cets_dist_blocker, which will set cookies that would prevent nodes from reconnecting before cleaning is done.

We use [set_cookie/2](https://www.erlang.org/docs/25/man/erlang#set_cookie-2), which allows to set a particular cookie for a specific remote node (and specific local node). If we set cookie to some random value, the node would not be able to connect to us (also we would not able to connect). But the existing connection will remain. Also, it only affects that specific node, other nodes would use cookie provided by `get_cookie/0` function.

The idea is:
- on nodeup we set cookie to something different for that particular node, so the next time it tries to reconnect, it will not able to.
- on nodedown we would wait for all the cleaner processes to ack that they've finished handling `nodedown` message (sent using `monitor_nodes` call).
- we only do it on a local node (i.e. no coordination between dist_blockers in cluster). Each MongooseIM node has a dist_blocker process though, just no messaging between them. Reason: simplicity. Also, that approach is enough to fix most of the issues.

This PR addresses "sessionCount is incorrect during upgrades".
The final goal is to make global's prevent_overlapped_partitions reliable.

Proposed changes include:
* cets_dist_blocker server, which provides API to block node from reconnecting, unless we do cleaning first.
* Test helper to start peer nodes.

PR to MongooseIM: https://github.com/esl/MongooseIM/pull/4234
PR to MongooseHelm: https://github.com/esl/MongooseHelm/pull/38

### DIFF
--- a/src/cets_dist_blocker.erl
+++ b/src/cets_dist_blocker.erl
@@ -67,7 +67,7 @@ handle_call({cleaning_done, CleanerPid, Node}, _From, State) ->
     {reply, ok, maybe_unblock(State, handle_cleaning_done(CleanerPid, Node, State))};
 handle_call(Request, _From, State) ->
     ?LOG_ERROR(#{what => unexpected_call, msg => Request}),
-    {reply, ok, State}.
+    {reply, {error, unexpected_call}, State}.
 
 handle_cast(Msg, State) ->
     ?LOG_ERROR(#{what => unexpected_cast, msg => Msg}),
--- a/src/cets_dist_blocker.erl
+++ b/src/cets_dist_blocker.erl
@@ -3,7 +3,7 @@
 %% This module prevents a node from reconnecting, until cleaning activity is
 %% finished. It prevents race conditions.
 %%
-%% This module assume all nodes share the same cookie.
+%% This module assumes all nodes share the same cookie.
 -module(cets_dist_blocker).
 -behaviour(gen_server).
 -include_lib("kernel/include/logger.hrl").
@@ -50,7 +50,8 @@ add_cleaner(CleanerPid) ->
 
 %% @doc Confirm that cleaning is done.
 %%
-%% This function is called by a cleaner after it receives nodedown.
+%% This function should be called by a cleaner when it receives
+%% nodedown and finishes cleaning.
 -spec cleaning_done(pid(), node()) -> ok.
 cleaning_done(CleanerPid, Node) ->
     gen_server:call(?MODULE, {cleaning_done, CleanerPid, Node}).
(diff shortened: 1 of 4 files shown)


src/cets_dist_blocker.erl
```diff
@@ -0,0 +1,156 @@
+%% @doc Disallow distributed erlang connections until cleaning is done.
+%%
+%% This module prevents a node from reconnecting, until cleaning activity is
+%% finished. It prevents race conditions.
+%%
+%% This module assume all nodes share the same cookie.
```
chrzaszcz:
```suggestion
%% This module assumes all nodes share the same cookie.
```

src/cets_dist_blocker.erl
```diff
@@ -0,0 +1,156 @@
...
+    gen_server:start_link({local, ?MODULE}, ?MODULE, [], []).
+
+%% @doc Register CleanerPid as a cleaner.
+-spec add_cleaner(pid()) -> ok.
+add_cleaner(CleanerPid) ->
+    gen_server:call(?MODULE, {add_cleaner, CleanerPid}).
+
+%% @doc Confirm that cleaning is done.
+%%
+%% This function is called by a cleaner after it receives nodedown.
```
chrzaszcz: This comment is a bit confusing, because it might suggest to call it right after nodedown.

chrzaszcz (commented):
Looks good. My only concern is how to protect against the cleaner taking too long on some node. Some timeout maybe? But let's not make it too complex at once.

arcusfelis:
> Looks good. My only concern is how to protect against the cleaner taking too long on some node. Some timeout maybe? But let's not make it too complex at once.

Could be timeout. Could be a very noisy logging. With timeout allowing node to connect would lead to the "unusual state" (i.e. untested state). It could be ok, it could be just adding more load to the cluster and causing errors.

In this case there is a chance that there would be "that one node, that prevented the join".
But it is fine-tuning, core functionality is here.

Currently we have:
- nodes are getting cleaned one-by-one.
- cleaning would take some reasonably short amount of time. Though not always, we have code that calls mod_last to update last_seen timestamp from SM on cleanup.
- there is a chance that node would be able to connect to one of the nodes that finished cleaning, and not connect to other nodes. Some coordination could be required. Could make a small test and an issue later on, but it is a separate issue.
- The idea is to have the basic dist_cleaner, and then maybe we can add cross-node cleaners (i.e. node would set cookie to unblocked on all nodes at once, if at least one node sets cookie to blocked - all nodes in the cluster would set to blocked).

Oh, I think there is 30 seconds grace period during which cets_discovery would not try to reconnect - so, it should be enough for the cluster to be able to accept the node again. Maybe it could be enough.

chrzaszcz (approved):

PR #53, opened 2024-03-12
Split tests

arcusfelis:
This PR addresses MIM-2187.

Separate cets_SUITE into parts

cets_SUITE is getting big, and big files are hard to understand/maintain.

- Create helper modules. Put generic code there.
- Split test cases into separate suites. Modules cets_join, cets_discovery and cets (contains insert/update operations)  should be tested by  different suites.


test/cets_SUITE.erl
```diff
@@ -12,6 +12,62 @@
...
+]).
+
+-import(cets_test_receive, [
+    receive_message/1,
+    receive_message_with_arg/1
+]).
+
+-import(cets_test_helper, [
+    assert_unique/1,
+    set_other_servers/2,
```
gustawlippa: I think this is not used in this suite.

test/cets_SUITE.erl
```diff
@@ -73,53 +121,11 @@ cases() ->
...
-        join_fails_before_send_dump_and_there_are_pending_remote_ops,
-        send_dump_fails_during_join_because_receiver_exits,
-        join_fails_in_check_fully_connected,
-        join_fails_because_join_refs_do_not_match_for_nodes_in_segment,
-        join_fails_because_pids_do_not_match_for_nodes_in_segment,
-        join_fails_because_servers_overlap,
-        remote_ops_are_ignored_if_join_ref_does_not_match,
-        join_retried_if_lock_is_busy,
         send_dump_contains_already_added_servers,
         servers_remove_each_other_if_join_refs_do_not_match_after_unpause,
```
gustawlippa: Just to be sure, is this not a testcase that could go to `cets_join_SUITE.erl`? It's the only one using the `set_join_ref/2` helper outside of this suite.

test/cets_test_helper.erl
```diff
@@ -0,0 +1,37 @@
...
+    sys:replace_state(Disco, fun(#{nodedown_timestamps := Map} = State) ->
+        State#{nodedown_timestamps := maps:put(Node, NewTimestamp, Map)}
+    end).
+
+set_other_servers(Pid, Servers) ->
+    sys:replace_state(Pid, fun(#{other_servers := _} = State) ->
+        State#{other_servers := Servers}
+    end).
+
+set_join_ref(Pid, JoinRef) ->
```
gustawlippa, line 36: Maybe if one testcase is moved from `cets_SUITE` to `cets_join_SUITE` this could be moved to `cets_join_SUITE` as well?

test/cets_test_helper.erl
```diff
@@ -0,0 +1,37 @@
+-module(cets_test_helper).
```
gustawlippa, line 1: I must admit I'm not a fan of a general helper, especially if there are specific helpers, but I'm not sure how to fix this.
I've left some comments which may make sense if we reshuffle some testcases, but I also understand if we want to move helper functions out of the `*_SUITE` files altogether - then this module is needed as it is I believe.

test/cets_join_SUITE.erl
```diff
@@ -0,0 +1,646 @@
...
+    simulate_disco_restart/1,
+    make_signalling_process/0,
+    given_two_joined_tables/1,
+    given_two_joined_tables/2,
+    given_3_servers/1,
+    given_3_servers/2,
+    given_n_servers/3
+]).
+
+-import(cets_test_wait, [
```
gustawlippa: I don't think these are used in this suite.

test/cets_netsplit_SUITE.erl
```diff
@@ -0,0 +1,259 @@
...
+
+-import(cets_test_setup, [
+    setup_two_nodes_and_discovery/1,
+    setup_two_nodes_and_discovery/2,
+    simulate_disco_restart/1,
+    make_signalling_process/0,
+    given_two_joined_tables/1
+]).
+
+-import(cets_test_wait, [
```
gustawlippa: I don't think these are used in this suite.

test/cets_status_SUITE.erl
```diff
@@ -0,0 +1,388 @@
...
+    wait_till_test_stage/2
+]).
+
+-import(cets_test_setup, [
+    setup_two_nodes_and_discovery/1,
+    setup_two_nodes_and_discovery/2,
+    simulate_disco_restart/1
+]).
+
+-import(cets_test_wait, [
```
gustawlippa: Similarly to netsplit and join suites, I think this could be removed?

test/cets_test_helper.erl
```diff
@@ -0,0 +1,37 @@
...
+-include_lib("eunit/include/eunit.hrl").
+
+-export([
+    get_disco_timestamp/3,
```
gustawlippa, line 5: This could be moved to the disco_SUITE along with `wait_for_disco_timestamp_to_be_updated` from the `cets_test_wait` helper, as it's the only place that uses this outside of the disco_SUITE... but I'm not really sure if moving the wait helper to the suite is a good idea, just some food for thought.

test/cets_test_helper.erl
```diff
@@ -0,0 +1,37 @@
...
+    #{MapName := #{NodeKey := Timestamp}} = Info,
+    Timestamp.
+
+%% Fails if List has duplicates
+assert_unique(List) ->
+    ?assertEqual([], List -- lists:usort(List)),
+    List.
+
+%% Overwrites nodedown timestamp for the Node in the discovery server state
+set_nodedown_timestamp(Disco, Node, NewTimestamp) ->
```
gustawlippa, line 26: This seems to be used in one place in disco_SUITE. I'd rather move it there, but I'm also fine with keeping the suites clean from helpers if we prefer this style.

test/cets_test_peer.erl
```diff
@@ -1,16 +1,61 @@
...
+    disconnect_node/2,
+    disconnect_node_by_name/2
+]).
+
+-import(cets_test_rpc, [rpc/4]).
+
 -include_lib("common_test/include/ct.hrl").
 
-start_node(Sname) ->
+start(Names, Config) ->
```
gustawlippa: Minor, but this argument is called an `Id` in other places, like the `find_or_start_node/1` function. From what I see, it's used more like and ID, so I'd go with this name for this argument everywhere.

test/cets_test_peer.erl
```diff
@@ -23,20 +68,46 @@ node_to_peer(Node) when Node =:= node() ->
...
+    rpc(Peer, erlang, set_cookie, [node(), erlang:get_cookie()]),
+    erlang:set_cookie(Node, erlang:get_cookie()),
+    %% Very rarely it could return pang
+    cets_test_wait:wait_until(fun() -> rpc(Peer, net_adm, ping, [node()]) end, pong),
+    cets_test_wait:wait_until(fun() -> rpc(node(), net_adm, ping, [Node]) end, pong).
+
+disconnect_node(RPCNode, DisconnectNode) ->
+    rpc(RPCNode, erlang, disconnect_node, [DisconnectNode]).
+
+disconnect_node_by_name(Config, Id) ->
```
gustawlippa: I think it would be better if this were called `disconnect_node_by_id`?

test/cets_netsplit_SUITE.erl
```diff
@@ -0,0 +1,259 @@
...
+-import(cets_test_receive, [
+    receive_message/1,
+    flush_message/1
+]).
+
+-import(cets_test_peer, [
+    block_node/2,
+    reconnect_node/2,
+    disconnect_node/2,
+    disconnect_node_by_name/2
```
gustawlippa: I think this is unused

test/cets_join_SUITE.erl
```diff
@@ -0,0 +1,646 @@
...
+    receive_message/1,
+    receive_message_with_arg/1,
+    flush_message/1
+]).
+
+-import(cets_test_peer, [
+    block_node/2,
+    reconnect_node/2,
+    disconnect_node/2,
+    disconnect_node_by_name/2
```
gustawlippa: I think this is unused.

test/cets_status_SUITE.erl
```diff
@@ -0,0 +1,388 @@
...
+
+-import(cets_test_receive, [
+    receive_message/1,
+    flush_message/1
+]).
+
+-import(cets_test_peer, [
+    disconnect_node/2,
+    disconnect_node_by_name/2
+]).
```
gustawlippa: I don't thing these are used. In general, there are some imports in the suites which are not used. It's not a big deal, but is there a reason for it?

gustawlippa (approved):
Thanks for the PR! Moves the suites to different suites in a sensible manner, and introduces relevant helpers, it's a great effort!
I've left some minor comments, mostly about unused import statements, and maybe moving some helper functions around.
I didn't really look into the test logic, but from what I see the tests are mostly moved around and the new changes make sense to me 👍

test/cets_test_helper.erl
arcusfelis: I like suites contain test cases, helpers could go into a separate file

gustawlippa: OK, I don't mind that. 👍 
I just worry a bit with a general test helper, it may contain everything but the kitchen sink after some time, but it also makes sense to keep the suites clean.

